PR #62, opened 2015-01-08
Add UI to share articles

shtrom:
This adds a menu option to share the article current being read through a
standard Android Intent.

The share option will show up in the ActionBar if space is available.

The shared form is similar to that of the webapp, with title in the subject, and
“URL via @wallabagapp” in the body of the intent.


tcitworld:
Seems good. I leave it to @kevmeyer.

kevmeyer:
Here I am, sorry for the delayed answer.  
Ok. I'm almost happy it ;)  
- could you remove the unused function `addShareButtonToActionBar` (this is done in our base class `BaseActionBarActivity`)  
- line 152: typo in "via @wallabagapp" (see https://github.com/shtrom/android-app/commit/a7684ef5707cbdfb8265cd98fd3f51d7e356abdb#diff-90817f28f871eb773dc83e2ecf7fca4eR132)  
- share menu icon: we could also use the system icon `@android:drawable/ic_menu_share` instead of shipping our own menu share-icon (or am I missing something here?)  

I'll merge it then.

shtrom:
> share menu icon: we could also use the system icon @android:drawable/ic_menu_share instead of shipping our own menu share-icon (or am I missing something here?)

Ah, right. I was surprised to have to add such basic pixmaps into the package, but I didn't see anything referring to system-wide images in the doc. This makes more sense.

All fixed now.

shtrom:
Ok, merged with the tip of dev branch. See comments in the merge commit for details. It build and works with version 21.0.0 of the build tools, and I do have both the mark as read and share buttons in the action bar.

kevmeyer:
ok, very nice.  
could you resend this PR to the dev-branch? I can merge it then.

PR #63, opened 2015-01-08
Add Poche::getDatabase to prevent overwriting reference, leading to a leak

shtrom:
NOTA: This is not thoroughly tested, but this seems to fix the problem, while not causing any other.

Fixes the following exception

    java.lang.IllegalStateException: /data/data/fr.gaulupeau.apps.InThePoche/databases/article_db.sqlite SQLiteDatabase created and never closed

Signed-off-by: Olivier Mehani <shtrom@ssji.net>


tcitworld:
Can you resend on dev branch ? Let's make things clean. :)

shtrom:
Sure. Should I also make one for the legacy branch?

tcitworld:
> Sure. Should I also make one for the legacy branch?

If I don't screw things up, merging dev into legacy should do it afterwards.

PR #66, opened 2015-01-12
Share intent adds site directly without opening browser, supports HTTP Auth too

moparisthebest:
This solves issue #9 , it also brings in the Jsoup library which could potentially be used for feeds as well.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -13,7 +13,7 @@ android {
     }
     buildTypes {
         release {
-            runProguard false
+            minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
     }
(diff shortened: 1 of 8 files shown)


tcitworld:
(Note : I didn't look through the code much and I don't know much to android)

This can pass though the login form ? If it does, it's quite an amazing work. You will be so much thanked !

Do you know if we could get the html response (logged in/linked saved/invalid password) and show it in the app ?
Also, Travis CI seems to show errors, but is it important ?

tcitworld:
>  Also, Travis CI seems to show errors, but is it important ?

Not related to you, sorry.

moparisthebest:
This doesn't deal with the login form because it sends user/pass with HTTP Authentication which wallabag supports, it does still have to deal with the craziness of logging in, getting a cookie, and then sending that cookie with the add request, but it works.

As far as the HTML response, it sends back an entire page worth, so rather than parse that out, I simply display (in a Toast) 'Success' if it returned 200 or 'Error' if something bad happened, and if I have a non-200 error code like 401 Unauthorized 'Error 401'.

tcitworld:
Can you resend your PR on dev branch, please ?

Issue #70, opened 2015-01-18
UI: Display list of articles by default, move front buttons to menu


shtrom:
Once configured, one will more generally start the app to read articles. The front screen with the buttons is therefore a bit useless, and delays accessing the articles.

I think it would make sense to move these buttons to the soft menu of the article list (maybe with the sync button in the ActionBar), and present that first.

kevmeyer:
I agree with you.  
Maybe we could also use a pull-to-refresh functionality to trigger the sync, if everyone is happy with it.

tcitworld:
I totally with you two.

>  Maybe we could also use a pull-to-refresh functionality to trigger the sync, if everyone is happy with it.

Also, automatic sync when connection settings are saved. :)

shtrom:
:+1: to Pull-to-refresh (:

tcitworld:
Pull To Refresh is available. And the rest seems ok.

PR #72, opened 2015-01-19
use article list as root activity

kevmeyer:
I started to work on #70. It's not done yet, but the basic functionality should be ready.  
Currently there's only a reload button (without any activity-indicator) instead of pull-to-refresh, but I think it's good for testing already.


kevmeyer:
Added pull-to-refresh.  
Feedback very welcome ;)

device-2015-01-22-164213

tcitworld:
I did test before you added P2R, will test this tomorrow or later.

PR #84, opened 2015-06-08
High contrast

shtrom:
Just so this is not lost: This PR tweaks the article CSS a bit to force high-contrast colours. This is useful on e-ink readers such as the Nook, where text is otherwise a bit hard to read sometimes. This is done by adding the relevant CSS class to the article.

At the moment, this is done only if a Nook is detected. This should probably be done in the settings somewhere, or at list with a list of known devices.

So, not a very functional patch at this point (though it doesn't break anything), but it can prove a useful thing to have in the future.

### DIFF
--- a/app/src/main/assets/main.css
+++ b/app/src/main/assets/main.css
@@ -16,6 +16,7 @@ html {
 
 body {
   background: #EEE;
+  font-family: serif;
 }
 
 .hicontrast {
--- a/app/src/main/assets/main.css
+++ b/app/src/main/assets/main.css
@@ -18,6 +18,11 @@ body {
   background: #EEE;
 }
 
+.hicontrast {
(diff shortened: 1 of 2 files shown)


tcitworld:
Did you or are you able to test it on your Nook ?

shtrom:
On Mon, Jun 08, 2015 at 07:38:23AM -0700, Thomas Citharel wrote:
> Did you or are you able to test it on your Nook ?

Yes, it has been working well for a while (:

-- 
Olivier Mehani <shtrom@ssji.net>
PGP fingerprint: 4435 CF6A 7C8D DD9B E2DE  F5F9 F012 A6E2 98C6 6655
Confidentiality cannot be guaranteed on emails sent or received unencrypted.

PR #90, opened 2015-07-23
updated some strings so they sound nicer

axelsimon:
A few updates to the text, to make it more obvious to new users. Also updated the copyright mention.


tcitworld:
thanks !

PR #96, opened 2015-10-17
Scroll position restoration in ReadArticle

di72nn:
Store/restore scroll position in/from local DB.

Restoring position after WebView's loaded is kinda [tricky](http://stackoverflow.com/a/7011229).

Position may be not so accurate on screen orientation change (probably because of text realignment).


di72nn:
This branch is obsolete.

PR #98, opened 2015-10-20
Major update

victorhaggqvist:
So.. the initial goal of this was to make selecting articles as read sync back to the server.

Then things got kind of hairy. Wallabag's article id fro the server was not present in the local DB. DB interactions is all handcrafted.

To make things easier to work with and since there has been no big updates in a while, I decided to make some changes. These changes are breaking, but that should be no big deal since everything lives on the server anyway, just resync.

To make DB things less manual, greenDAO is used to handle all that, which makes things less easy to break by mistake. The FeedUpdater have been extended to store article ids, not just the internal id.

To make it possible to actually sync selecting articles as read, username and password is now settings to. Since that is done by first signing in and then issue a request to `/?action=toggle_archive&id=<article_id>`. To make this less painful OkHttp is now used instead of the regular HTTP client. As a result of now having login info, adding articles is now automated also, instead of the browser bounce.

Also, to separate code, the share action is now broken out to BagItProxyActivity.

I might have missed some things here, but most of the transitioning things are made in https://github.com/wallabag/android-app/commit/3dd80c032cd7eae74eac149c8208d89b9774219d.

This is a changeset against the master branch. While writing this I also just realized that there was a dev branch with a few changes. Those would probably be pretty easy merged in to this as well.


tcitworld:
Wow. This look like real great stuff. I'll do a build and merge asap.

However, I need to warn you that we've worked to release a v2 version of wallabag that provides [an OAuth-based API](https://www.wallabag.org/blog/2015/10/06/developers-oauth2) that provides [basically everything we need](http://v2.wallabag.org/api/doc). There's also an (now outdated) PoC for a [new ionic android app](https://github.com/wallabag/ionic-app). Just so you don't waste your time.

But your work is truly welcomed anyhow !

victorhaggqvist:
@tcitworld Happy to hear that :)

I do know about v2, looking forward to see how it turns out.

di72nn:
I was making some improvements for FeedUpdater recently: rewrote using XmpPullParser, added processing of "Archive" and "Favorites" feeds, added somewhat hackish "fast" sync. Not so obvious API, though. If it may be useful: https://github.com/di72nn/android-app/tree/sync_improvement

tcitworld:
> I may incorporate my changes here.

You're very welcomed to send a PR, based on this may be better.

di72nn:
Implemented my changes on top of victorhaggqvist's branch [here](https://github.com/di72nn/android-app/tree/sync_improvement_for_m-u).

With the old code it took about 45 seconds to update "home" feed on my device, with my code it took about 30 seconds to update "home" + "fav" + "archive" feeds. It also can load a couple of new articles from "home" feed in seconds (though, it's not yet available in UI; also UI may be updated to view or to prioritize "favorite" articles).

Not sure if I should create a separate PR.

PR #119, opened 2015-11-13
Swipe views and some other updates

di72nn:
* Swipe views for "Unread", "Favorite" and "Archived" lists.
* A little update for prev/next buttons: thee buttons now honor context (e.g. if article was opened in "favorites" list, the buttons will move though articles in "favorites" list (well, I hope they would)).
* Minor changes.

Full/fast sync explanation:
Well, there is no actual *synchronization*, only one-way *update*.
"Update all feeds" option in menu starts full update (wipes DB, successively updates "Unread", "Archived" and "Favorites" feeds). There are no more other update options in menu.
In each list (in the mentioned swipe views) pull-to-refresh starts update for the corresponding feed.
For "Favorite" and "Archived" feeds it's pretty easy: read all feed and update local entries.
For "Unread" feed update has a little hack: it reads feed and adds new entries, but stops as soon as it encounters an article with the same articleId it has locally. It allows quickly load recently added articles (though, it won't load some old moved-from-archive article). This is previously so called "Fast sync". This also is the default behavior for pull-to-refresh for "Unread" feed. I probably should make it optional.


tcitworld:
Just two things : 
* When you open the app, the favorite text tab is selected (bold) while it's the Unread panel which is focused and showed.
* If I go to the first article of any category, it's the wrong button (previous instead of next) that is shown and works.

Since you bring up this new display, I'll delete the fav and archive status icons I had brought on each item of the list of articles.

Thanks again for your awesome work !

di72nn:
Fixed.

I'm not sure status icons should be removed altogether: "Unread" and "Archived" can have stars, "Favorites" can have archive icons. I can rewrite ListAdapter to show only needed icons for different lists, if that is desired behavior.
Though, I experience weird behavior: for some reason archive icons don't appear right after I open "Favorites" tab, but do appear after I scroll up and down. Not sure about other feeds.

tcitworld:
Seems good now.

>  Though, I experience weird behavior: for some reason archive icons don't appear right after I open "Favorites" tab, but do appear after I scroll up and down. Not sure about other feeds.

Yup, had that too once. Maybe it's will be fixed if I use icons instead of unicode strings.

Issue #118, opened 2015-11-13
[1.9] Can't sync


khinsen:
I just signed up for the beta testing program and installed the new app. Then I configured it for my Framabag account. Result: "Erreur de traitement du flux" (my phone is configured in French).

Of course I checked all my configuration info (login, password, user ID, token) ten times. I also regenerated a new token and double-checked it. No chance.

In the "settings" dialog, the connection test works fine, but the one for obtaining the RSS info leads to an error message.

m-dubois:
Same for me here, but using my own wallabag server.
Do yo need more infos?

m-dubois:
C'est un peu le m*rdier dans logcat :p
Mais j'ai quand même vu la stack suivante qui pourrit p'têt aider:

```
E/FeedUpdater.updateByF( 2904): XmlPullParserException on Main
E/FeedUpdater.updateByF( 2904): org.xmlpull.v1.XmlPullParserException: expected: START_TAG {null}rss (position:START_TAG <html class='ng-csp' data-placeholder-focus='false' lang='en'>@7:82 in java.io.InputStreamReader@41eaaf78) 
E/FeedUpdater.updateByF( 2904): 	at org.kxml2.io.KXmlParser.require(KXmlParser.java:2056)
E/FeedUpdater.updateByF( 2904): 	at fr.gaulupeau.apps.Poche.network.tasks.FeedUpdater.processFeed(FeedUpdater.java:226)
E/FeedUpdater.updateByF( 2904): 	at fr.gaulupeau.apps.Poche.network.tasks.FeedUpdater.updateByFeed(FeedUpdater.java:173)
E/FeedUpdater.updateByF( 2904): 	at fr.gaulupeau.apps.Poche.network.tasks.FeedUpdater.updateAllFeeds(FeedUpdater.java:107)
E/FeedUpdater.updateByF( 2904): 	at fr.gaulupeau.apps.Poche.network.tasks.FeedUpdater.doInBackground(FeedUpdater.java:71)
E/FeedUpdater.updateByF( 2904): 	at fr.gaulupeau.apps.Poche.network.tasks.FeedUpdater.doInBackground(FeedUpdater.java:30)
E/FeedUpdater.updateByF( 2904): 	at android.os.AsyncTask$2.call(AsyncTask.java:288)
E/FeedUpdater.updateByF( 2904): 	at java.util.concurrent.FutureTask.run(FutureTask.java:237)
[...]

PR #120, opened 2015-11-14
Updates

di72nn:
* Fix icons appearing/disappearing.
* Show relevant icons for each list.
* Improve "connection test" due to issues like #118


tcitworld:
Nice. However, we need to change the unicode icons to pictures, it seems some devices do not support them. My fault, sorry.

tcitworld:
Perfect. :)

Issue #122, opened 2015-11-15
navigate between articles


nicosomb:
When I'm reading an article, I want to swipe to read the next article.

tcitworld:
This may be tricky since content is displayed in a webview.

nicosomb:
Why do we use webview?

tcitworld:
Hmm, it's quite the best option we have to display stylised text. Allows also YouTube videos to be read.

nicosomb:
Can you have a look to the Reflets.info app ? Is that possible to implement this way of view and swipe ?

di72nn:
[The app](https://github.com/freebyte/Reflets-android/blob/master/src/info/reflets/app/ArticleActivity.java) uses a couple of predefined views (title, date, author, etc.) and generates main content view via [Html.fromHtml](https://developer.android.com/reference/android/text/Html.html#fromHtml%28java.lang.String,%20android.text.Html.ImageGetter,%20android.text.Html.TagHandler%29). I believe WebView is more suitable currently.

If I understand correctly, "swipe" is already implemented (*more or less*).

nicosomb:
Yes, swipe is already implemented, but it's not very convenient to use.

di72nn:
Sorry to duplicate (issues)[#134], but this one leads me into wrong direction :)

Issue #20, opened 2014-05-30
HTTP authentication support


pmaziere:
HTTP authentication is currently missing which makes this application useless for anyone hosting its wallabag with privacy in mind 

thanks

EtiennePerot:
+1. I don't trust random self-hosted web apps to be secure. I want nginx's auth on top of it for extra safety. tt-rss's Android app for example lets me set the HTTP username/password exactly for this reason.

tcitworld:
wallabag v2 doesn't support HTTP auth for now, please open a new issue there if you are still interested with this feature. I close here for now.

EtiennePerot:
:(

tcitworld:
This doesn't mean we are closed to the idea, just that we focus on other things for now.

tcitworld:
HTTP Auth is now available, thanks to @di72nn \o/

di72nn:
HTTP Auth needs testing. For details see the first comment at #124

tcitworld:
The latest build is available on Google Play for testers : https://play.google.com/apps/testing/fr.gaulupeau.apps.InThePoche

nicosomb:
New beta?

On dim., nov. 15, 2015 at 7:24 PM, Thomas Citharel < notifications@github.com [notifications@github.com] > wrote:
The latest build is available on Google Play for testers : https://play.google.com/apps/testing/fr.gaulupeau.apps.InThePoche
[https://play.google.com/apps/testing/fr.gaulupeau.apps.InThePoche]

—
Reply to this email directly or view it on GitHub
[https://github.com/wallabag/android-app/issues/20#issuecomment-156841505] .[https://github.com/notifications/beacon/AAHcDpaOHrOk7QEOxHQM6QHGVV9gZx-vks5pGMVSgaJpZM4B_tyW.gif]

tcitworld:
Yup.

nicosomb:
Not available for me

On dim., nov. 15, 2015 at 7:51 PM, Thomas Citharel < notifications@github.com [notifications@github.com] > wrote:
Yup.

—
Reply to this email directly or view it on GitHub
[https://github.com/wallabag/android-app/issues/20#issuecomment-156842933] .[https://github.com/notifications/beacon/AAHcDg-53qPKvBNo2m1afni1oVqGo4qqks5pGMu8gaJpZM4B_tyW.gif]

tcitworld:
[...]

PR #124, opened 2015-11-15
Updates

di72nn:
* Add Basic HTTP Auth support. #20 
* Add switching between articles with "fling" action. #122 

I'm pretty sure HTTP Auth won't break anything, but I'm also **not** sure it currently works (feeds + network commands will probably work; web-content (images from server) may not work (the code to support it is added, but not debugged)).

Swipe/fling action works, but may need some tuning (especially on screens with different size/density).

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/App.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/App.java
@@ -7,6 +7,7 @@ import com.facebook.stetho.Stetho;
 import fr.gaulupeau.apps.InThePoche.BuildConfig;
 import fr.gaulupeau.apps.Poche.data.DbConnection;
 import fr.gaulupeau.apps.Poche.data.Settings;
+import fr.gaulupeau.apps.Poche.network.WallabagConnection;
 
 /**
  * @author Victor Häggqvist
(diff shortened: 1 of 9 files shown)


di72nn:
N.b.: The swipe/fling **may** break some touch-related actions in WebView (unlikely) - unusual behavior should be reported.
There are log messages with "FLING" tag that can be used to calibrate swipe/fling action (essentially, they just show why swipe/fling wasn't considered valid; if you did swipe/fling, but nothing appeared in log - `GestureDetector` didn't considered it as a fling at all (probably harder to calibrate)).

tcitworld:
Yup, I just had an issue where there's a table inside content that has a width forced bigger than the screen, it blocks swipe on right side.


I've have however quite a few difficulties with some articles, and it was much easier with those values, but they may lead to problems : 
```
if(Math.abs(velocityX) < 80) {
                    Log.d("FLING", "too slow");
                    return false; // too slow
  }

if(Math.abs(velocityX / velocityY) < 3) {
                    Log.d("FLING", "not a horizontal fling");
                    return false; // not a horizontal fling
                }
```

tcitworld:
Note : My touchscreen is 1080*1920.

Issue #108, opened 2015-11-12
Dark theme


tcitworld:
For reading at night

tcitworld:
cc @Caliandroid for this one

Caliandroid:
Done:  https://github.com/wallabag/android-app/commit/b1591558213f3e036defe7e332fe2a3542e0e88f

Strubbl:
Isn't this already implemented? Okay, maybe not in the About dialog, but when reading articles. Can be closed?

tcitworld:
Yup. The About dialog style can be done easily with https://github.com/mikepenz/AboutLibraries/wiki/HOWTO:-Style.

Closing here.

PR #132, opened 2015-11-16
 Dark theme #108 added

Caliandroid:
The commit b159155 contains the dark theme/night view #108 

commit 8a3fca3 should be ignored (only a gradle.build update).
I've added my app.iml -> should be ignored too


Caliandroid:
I hadn't installed the gitignore plugin in android studio, that is why app.iml was not ignored

tcitworld:
Thanks. I just made a modification for it to work also on list articles.

tcitworld:
> I just made a modification for it to work also on list articles.

My bad, it's just it doesn't refresh.

tcitworld:
I've made a workaround to avoid using `recreate()` and refresh without restarting the app.

Caliandroid:
Thank you too for workaround.

PR #141, opened 2015-11-20
Updates

di72nn:
* Better themes support (all activities; theme selection from list; better colors).
* Serif typeface for article fonts as an option (was: inconsistent, depending on main/nightmode.css).
* Font size for articles fonts option.

I'm not into CSS, so improvements are very welcome.
Font setting implementation is plain stupid, but, well, you are free to replace it anytime.

Update: icons color for dark themes **was not** fixed in this update.


nicosomb:
Waiting for the new alpha! Really love this new app.

tcitworld:
>  Font setting implementation is plain stupid, but, well, you are free to replace it anytime.

The best would be something like what Pocket does, just a +/- setting with preview. But for now, it's quite fine.


However, we've got to change the size of the spinner item, they are quite small to click on :

nicosomb:
the 1st screenshot is a pocket one?

tcitworld:
Yup, don't get your hopes to high for now.

nicosomb:
so sad ;-)

PR #142, opened 2015-11-21
Experimental updates

di72nn:
Added options to increase/decrease font size and change font family to ReadArticleActivity options menu.
The change of these parameters implemented via JavaScript calls. Unfortunately, after every change we have to adjust reading position (that was implemented the same way as position restoration on article load, since I couldn't find any new hook).
All of that is kind of experimental technique.

Two points worth mentioning:
* Initial font size and font family are applied **after** the article is loaded (I'm not into HTML+CSS, so I don't know if there is a robust way to set it before loading (**and** don't break the ability to change these parameters after loading)).
* I have no idea how to adjust font parameters correctly, so I just change style on "body" element. I presume somebody else can fix that.


tcitworld:
> change font family

I think this is a setting that people choose one and for all, no need to put it in the ReadArticleActivity option menu, settingsActivity is enough.

I tried to apply the css transition property `document.getElementById("html-base-body").style.transition = "all 0.5s";` to make the changing size operation more fluid, but it doesn't makes things better.

However, I did find that you can change font size for the whole webview, did you give it a try ? 
`webSettings.setTextSize(WebSettings.TextSize.SMALLEST);` or `webSettings.setDefaultFontSize(10);`
There's also `settings.setTextZoom(settings.getTextZoom() + 10);` that exists.

tcitworld:
I submitted you https://github.com/di72nn/android-app/pull/2

di72nn:
Thanks for the suggestion. Sometimes I get weird ideas (using JavaScript in this case) that I can't get out of my head. Merged and removed JS altogether.
Also removed "serif" option from ReadArticleActivity (left only in Settings).

tcitworld:
No worries. Thanks for work.

PR #161, opened 2015-12-07
Fix Russian translation

glesik:
Confused app author for translation author. :)


tcitworld:
We will add an about page for contributors and translators later. :)

Issue #169, opened 2015-12-13
Another NullPointerException on WallabagService.getConfigRequest


tcitworld:
If url isn't set up ?

```
java.lang.RuntimeException: An error occured while executing doInBackground()
	at android.os.AsyncTask$3.done(AsyncTask.java:300)
	at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:355)
	at java.util.concurrent.FutureTask.setException(FutureTask.java:222)
	at java.util.concurrent.FutureTask.run(FutureTask.java:242)
	at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
	at java.lang.Thread.run(Thread.java:864)
Caused by: java.lang.NullPointerException
	at fr.gaulupeau.apps.Poche.network.WallabagService.getConfigRequest(WallabagService.java:197)
	at fr.gaulupeau.apps.Poche.network.WallabagService.getCredentials(WallabagService.java:51)
	at fr.gaulupeau.apps.Poche.network.tasks.GetCredentialsTask.doInBackground(GetCredentialsTask.java:40)
	at fr.gaulupeau.apps.Poche.network.tasks.GetCredentialsTask.doInBackground(GetCredentialsTask.java:14)
	at android.os.AsyncTask$2.call(AsyncTask.java:288)
	at java.util.concurrent.FutureTask.run(FutureTask.java:237)
	... 4 more
```

PR #170, opened 2015-12-13
Fix NPE

di72nn:
#169. Also fixed another possible uncaught exception.

Oh, and I also removed `values-en` because it just duplicates `values`.


tcitworld:
> Oh, and I also removed values-en because it just duplicates values.

Sorry I not responded about this earlier. I don't know why, but the Google Play store console complained each time I uploaded an APK without this translation file. Don't really know why, so I guess keeping it is better, since it's not really a problem.

app/src/main/res/values/strings.xml
```diff
@@ -90,6 +90,7 @@ GAULUPEAU Jonathan — 2013-2015
     <string name="testConnection_errorMessage2">Can\'t login; check username and password. Error code: %d</string>
     <string name="testConnection_errorMessage3">Authorization-related problems. Error code: %d</string>
     <string name="testConnection_errorMessage5">Can\'t login; check HTTP Auth parameters. Error code: %d</string>
+    <string name="testConnection_errorMessage6">Illegal URL. Error code: %d</string>
```
tcitworld: You mean incorrect URL, right ?

di72nn:
"Illegal" -> "Incorrect".
Symlinked `values-en/strings.xml` to `values/strings.xml`. Let's see if that's enough.

PR #192, opened 2016-01-31
Add solarized-like theme

stepardo:
This adds a theme that is inspired (but not as good and complete) by the
solarized theme by Ethan Schoonover [0].

[0] http://ethanschoonover.com/solarized

This is my first attempt, and I am no expert in css/web programming. So please be patient with me.

### DIFF
--- a/app/src/main/assets/solarized.css
+++ b/app/src/main/assets/solarized.css
@@ -8,6 +8,23 @@
    5 = Article
    6 = Media queries
 
+   *solarized colors*
+$base03:    #002b36;
+$base02:    #073642;
+$base01:    #586e75;
+$base00:    #657b83;
+$base0:     #839496;
+$base1:     #93a1a1;
+$base2:     #eee8d5;
+$base3:     #fdf6e3;
+$yellow:    #b58900;
+$orange:    #cb4b16;
+$red:       #dc322f;
+$magenta:   #d33682;
+$violet:    #6c71c4;
+$blue:      #268bd2;
+$cyan:      #2aa198;
+$green:     #859900;
    ========================================================================== */
 
 html {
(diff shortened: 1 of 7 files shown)


tcitworld:
Thanks. Red links are a little shocking though. But the other accent colors are not quite satisfying either. Do you have another proposition ?

nicosomb:
Thanks for this PR :+1:

stepardo:
Ok, I can make the red links darker, which will help. And I guess the color of the android widgets is too strong?

Do you know how I can set the color of buttons?

Rant:
_The theme editor in android studio is a very good example for very bad UI...Also, I can set *every* color with a css-style hex value *except* windowBackground, which requires an indirection via colors.xml. Maybe its just me, but so far the experience of android development is not enjoyable._

stepardo:
I think I improved the theme a lot by using the actual colors from the solarized website. Please check it out. There are two glitches in the settings activity, where some text is still black, but I could not figure out why. Also, could somebody please test it with a device of API level 8 (lowest supported by wallabag). I tested with v21 and above.

stepardo:
Ok, rewrote it to have the correct colors as defined by the solarized theme. Please test;-)

tcitworld:
I like it. Can the action bar have a slighty different color than the rest of the window ?

stepardo:
On 03.02.2016 19:09, Thomas Citharel wrote:
> I like it. Can the action bar have a slighty different color than the
> rest of the window ?

In terms of solarized color codes it has base2, whereas the rest is 
base3. So it is slightly darker (you can see that if you open the menu. 
Solarized is all about weak contrasts.

tcitworld:
I can do with that. :+1:

PR #204, opened 2016-04-11
V2 (and V1) login

Strubbl:
These commits create the possibility to login against a Wallabag V2 *and* V1 version. 

That means, if you go into your settings and either enter a v1 or v2 valid wallabag URL and hit the "Test connection" button, it will work.


Strubbl:
closing due to PR #210 handling all v2 stuff

Issue #203, opened 2016-04-07
app V1.9.x incompatibility with Wallabag V2.0.x


fodiator:
I'm actual not shure if it is known that the current android app seems to be incompatible with the recently released V2.0 server-version?

Seems like a bug to me

Kind regards

GLLM:
Same for me !
THe URL to use is not mentioned anymore in the documentation !

giovi321:
Same problem here. I get the error "Wallabag service not found; check URL. Error code: 1".

gerroon:
I have the same issue

tcitworld:
Hi there. It's a work in progress at #204. Expect it at the end of the week.

Strubbl:
if you want to follow progress, #210 is the PR to watch for. With those changes basically v2 is supported now.

fodiator:
unfortunately V 1.9.9.1 doesn't work for me. Although wallabg server (2.0.1) is operable (and token has been created) the app responds "failed to connect to XXX"
Any idea?

Strubbl:
The check fails because the App cannot find the following string on your Wallabag URL:

`<form action="/login_check" method="post" name="loginform">`

Maybe that changed from 2.0.0 to 2.0.1. I don't have that version running. If you can provide the URL to your instance I can double-check it.

fodiator:
THX
please test it at https://ip.2hh.at:45443/wallabag

Strubbl:
Root cause for this problem is, that the wallabag instance is located in the subdirectory wallabag/, but the app expects it directly in /
That needs to be fixed in several places in the app.

fodiator:
OK, thanks
I need to have wallbag in a subdir because i have many different services running on my server
Will close the port now, please advice if you need to have it opened again

Strubbl:
No problem, but it would be nice if you can test again if PR #212 is merged.

fodiator:
Hi, I've just tested V1.9.9.2 and login (connection test)  is working now, and the right RSS token is catched.
Unfortunately Updating feeds results in a 404 not found message
Kind regards

tcitworld:
Did you make sure to save the settings after getting the RSS token ? There's a save button at the bottom of the settings that needs to be pressed so that the new settings must be applied.

fodiator:
RSS token settings are saved definately.
Shall I open a port for you again?

tcitworld:
If you can do so (I'll create an account).

fodiator:
opened

[...]

PR #212, opened 2016-04-15
fix the v2 check for valid Wallabag instance

Strubbl:
This PR fixes a problem reported in https://github.com/wallabag/android-app/issues/203#issuecomment-210520327

The problem occurs when Wallabag v2 is not installed to / path of the
webserver, but in a subdirectory, e.g. wallabag/. Our regex check for
a validwallabag instance does not work anymore. Therefore the regex has
been shortened in the beginning.


Strubbl:
There are more commits to come in this PR. The logout link check does not work, too

Strubbl:
Okay, i hope i have corrected all places where the subdirectory would have been required. All other places use the endpoint variable, which contains the full URL incl. subdir.

tcitworld:
Merging this for now, but in the future we should definitely check for `/api/version`.

Strubbl:
Oh yes, didn't know that we can access api with only being logged in. I thought we had to implement oauth.

PR #213, opened 2016-04-15
App widget unread article count

Strubbl:
I created a widget which shows how many articles are saved in the Wallabag database of your Android app, which are unread. So if you are limiting the RSS feed in your Wallabag v2 (rss feed limit < no. unread articles), you will only see that limit and not the real unread count you see in the web browser.

features of the widget:
- on click start the application
- trigger widget counter update on sync, archive or delete action in Wallabag app

_Thanks to K-9 Mail for inspiration._

### DIFF
--- a/app/src-gen/fr/gaulupeau/apps/Poche/entity/ArticleDao.java
+++ b/app/src-gen/fr/gaulupeau/apps/Poche/entity/ArticleDao.java
@@ -1,6 +1,7 @@
 package fr.gaulupeau.apps.Poche.entity;
 
 import android.database.Cursor;
+import android.database.DatabaseUtils;
 import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteStatement;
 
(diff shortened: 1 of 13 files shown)


tcitworld:
I just have an issue with aligning the widget with regular icons. The K9-Mail widget works properly.

Strubbl:
I cannot reproduce that at my Android. But @layout/icon_unread might be the problem.

Strubbl:
@tcitworld do you have an idea why the layout does not work at your phone? Do you know how to fix that?

Otherwise some more details about the android version would be nice. I can try to reproduce it in the emulator then.

tcitworld:
>  do you have an idea why the layout does not work at your phone?

I use CyanogenMod nightlies of Android 6.0.1 with Trebuchet (which is therefore also in nightly) so I guess it could be an issue with my device. You can test with an emulator to be sure.

Strubbl:
With an Android 6.0 device I can reproduce this issue.

Strubbl:
@tcitworld Can you please test again if my last commit fixed your issue. If yes, this PR is ready to be merged, isn't it?

tcitworld:
Perfect !

PR #214, opened 2016-04-17
guessWallabagVersion() makes use of /api for v2

Strubbl:
Refactores guessWallabagVersion() into several methods having just one
task. The isWallabagVersion2() makes use of the provided API which was
introduced since v2.
Needed to increase minSdkVersion for isEmpty() check on Strings. (When some day the Widget PR will be merged, the minSdkVersion is increased anyway.)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -6,7 +6,7 @@ android {
 
     defaultConfig {
         applicationId "fr.gaulupeau.apps.InThePoche"
-        minSdkVersion 8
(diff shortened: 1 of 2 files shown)


tcitworld:
Judging from the Google Play stats, we can even put the SDK min version up to 14 with so significant loss of users, so no worries.

tcitworld:
:+1: 
Will check tonight.

PR #215, opened 2016-04-17
Fixes getCredentials() if Wbg installed in subdir

Strubbl:
This commit fixes the problem with requesting the FeedsCredentials when
Wallabag v2 is installed in a subdirectory of the server e.g.
endpoint=http://server/wbg/. This is done with a check of the userID for
a slash. In the given example the userID would have been "wbg/user". Now, if a slash is found, only the string right of the slash ("user") is used as userID.

Hopefully a user name never contains a slash ;)

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagServiceEndpointV2.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagServiceEndpointV2.java
@@ -80,7 +80,11 @@ public class WallabagServiceEndpointV2 extends WallabagServiceEndpoint {
     }
 
     public FeedsCredentials getCredentials() throws IOException {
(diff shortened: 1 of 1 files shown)


tcitworld:
Not sure about this, but what if an user has a `/` inside it's username/userID ?

Strubbl:
I thought that was not possible, but i could register at framabag with a slash in my name. Okay, then we shouldn't merge this as of now. 

It is getting more complex now. I need to check the endpoint variable a third slash and take the string after that as a temporary var. Now this temp. var needs to be substracted from the userID which we got from the /config. 

Or we go for using the API for everything now.

Strubbl:
in case it's v2, the new commit uses the user name given by the App user

tcitworld:
> I thought that was not possible, but i could register at framabag with a slash in my name

It is possible on wallabag v2 too.

PR #248, opened 2016-05-04
Refactoring

di72nn:
Mostly minor refactoring. The only more or less significant change is lazy initialization of `serviceEndpoint` in `WallabagService` -- the constructor is sometimes called in UI thread, that should not wait for `guessWallabagVersion()` to finish.

BTW, we have to do something with wallabag version guessing: right now the app makes at least two additional calls to server before **every** request (mark read, favorite, etc.). We shouldn't do that on mobile phones (well, *especially* on mobile phones).
The most viable solution is to detect server version on saving settings, but it's not very convenient.
It is also possible to add "Wallabag server version" switch to the settings screen, which would be automatically set via 'Get credentials' button.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagService.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagService.java
@@ -14,6 +14,8 @@ import fr.gaulupeau.apps.Poche.data.FeedsCredentials;
 import fr.gaulupeau.apps.Poche.data.Settings;
 
 import static fr.gaulupeau.apps.Poche.network.WallabagConnection.getRequest;
+import static fr.gaulupeau.apps.Poche.network.WallabagServiceEndpointV1.WALLABAG_LOGIN_FORM_V1;
+import static fr.gaulupeau.apps.Poche.network.WallabagServiceEndpointV1.WALLABAG_LOGOUT_LINK_V1;
 
 /**
  * @author Victor Häggqvist
@@ -124,7 +126,7 @@ public class WallabagService {
         String body = getBodyFromHttpResponse(endpoint + "/?view=about");
         if (body != null
                 && !body.isEmpty()
-                && (body.contains(Settings.WALLABAG_LOGOUT_LINK_V1) || body.contains(Settings.WALLABAG_LOGIN_FORM_V1))
+                && (body.contains(WALLABAG_LOGOUT_LINK_V1) || body.contains(WALLABAG_LOGIN_FORM_V1))
                 ) {
             Log.d(TAG, "isWallabagVersion1() found Wallabag v1");
             return true;
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagService.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagService.java
@@ -28,22 +28,25 @@ public class WallabagService {
     private String endpoint;
     private OkHttpClient client;
     private String username, password;
-    private int wallabagMajorVersion;
+    private int wallabagVersion;
     private WallabagServiceEndpoint serviceEndpoint;
-    private Settings settings;
 
     public WallabagService(String endpoint, String username, String password) {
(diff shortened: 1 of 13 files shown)


nicosomb:
Maybe my question is stupid. 
But why don't have 2 applications for the 2 versions of wallabag?

tcitworld:
Well, it's rather confusing for users. However, that leaves a lot of legacy code.

app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagService.java
```diff
@@ -33,51 +36,60 @@ public WallabagService(String endpoint, String username, String password) {
...
+    }
+
+    private WallabagServiceEndpoint getServiceEndpoint() {
+        if(serviceEndpoint == null) {
+            if(this.wallabagMajorVersion == -1) {
+                this.wallabagMajorVersion = guessWallabagVersion();
+            }
+
+            switch (this.wallabagMajorVersion) {
+                case 2:
```
nicosomb: the `case` are not ordered. It's not a problem in Java?

di72nn: Nope, the switch works exactly as it should:
* if `wallabagMajorVersion == 2` it creates V2 endpoint,
* if `wallabagMajorVersion == 1` it creates V1 endpoint,
* otherwise it logs warning and falls into `wallabagMajorVersion == 1` case.

di72nn:
I do think that client for WallabagV2 should be separate (forked from the current one?). But I fear that V1 client will be abandoned, if V2 is forked now.

app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagService.java
nicosomb: yes, but the `default` is not at the end of your case. Not a problem?

di72nn: Not a problem.

app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagServiceEndpointV1.java
```diff
@@ -24,13 +23,18 @@
  * Created by strubbl on 11.04.16.
  */
 public class WallabagServiceEndpointV1 extends WallabagServiceEndpoint {
+
+    public static final String WALLABAG_LOGIN_FORM_V1 = "<form method=\"post\" action=\"?login\" name=\"loginform\">";
+    public static final String WALLABAG_LOGOUT_LINK_V1 = "href=\"./?logout\"";
```
Strubbl: good idea to move that constants to the corresponding service version. now we have all data in one place

app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagService.java
Strubbl: @nicosomb it's not a problem cause there is no break; in the default case. Thus it falls through to case 1. So if wallabagMajorVersion is anything but one or two, it also uses V1 endpoint.

@di72nn Shouldn't we make V2 the default, because it is the current release?

di72nn:
Ok, here's an update. I added "Wallabag server version" selection to SettingsActivity. I also made `Get credentials` button to automatically select detected server version (though, it's not obvious in UI).
Thereby all the "usual network actions" don't try to autodetect server version.

I don't think that's a good solution (mostly UI problem), but that's better that making to much requests.

Strubbl:
In the long term we should remove the credential and version options from the settings menu. Only URL, user, password should be required. The rest can be automated and should not bother the app user.

For now, lets merge those refactoring changes. I like that the number of request is reduced.

di72nn:
> In the long term we should remove the credential and version options from the settings menu.

Totally agree. Just wanted to make the app a bit more usable until we get to these improvements.

Strubbl:
Yes, i know. :) Thanks for your commits.

PR #256, opened 2016-05-16
workaround for okhttp not working with http2 on nginx

tcitworld:
Fix for https://github.com/square/okhttp/issues/2543

Taken from : https://github.com/wikimedia/apps-android-wikipedia/commit/058ed9cb5f4b77411f113cb31820e4a8687ee11a

Thanks to Wonderfall for reporting.


Strubbl:
Why is this assigned to me? Okay, i just reviewed and it is looking fine. But i cannot test it due to not having an http2 server.

PR #257, opened 2016-05-19
building as debug results in different app (id)

Strubbl:
This commit introduces a different package name for the debug build of
this application. It is useful when having installed Wallabag app from
an app store and besides developing the app. You can now have a debug
and release version installed on your device in parallel.
The debug version of this app has a different app id, a different version name and a different
launcher icon (same as release, but with a small green bug).


tcitworld:
Nice picture. :+1:

PR #281, opened 2016-06-27
Replace Android-internal icons with SVG-resources

di72nn:
It was bothering me because it is not forward compatible.


tcitworld:
I was about to push the same changes. ;-)

PR #289, opened 2016-09-04
Add new, update and rearrange strings

strobeltobias:
- Missing strings added.
- Old strings updated to be (grammatical) better understandable.
- Strings rearranged to match the string order of the main strings.xml (in English). Might better help to translate and add new strings in future.


nicosomb:
Thank you!

PR #297, opened 2016-09-19
add German translation strings, fix some English and a Spanish string

Strubbl:


Strubbl:
I just tested my translation and I am going to improve some of the strings I just saw live in the app.

PR #298, opened 2016-09-19
[WIP] Fix lint issues

Strubbl:
I have started to fix some lint issues. Everybody is welcome to review my changes and contribute commits to this branch (fix-lint-issues).


app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
```diff
@@ -746,7 +746,13 @@ private String readRawString(int id) throws IOException {
...
         } finally {
             if(reader != null) {
-                reader.close();
+                try {
+                    reader.close();
+                } catch (IOException e) {
+                    // unrecoverable exception, only log it and move on
+                    Log.w(TAG, "readRawString() failed to close reader\n" + e.toString()
+                            + "\n" + e.getStackTrace().toString());
+                }
```
di72nn: We close reader for an internal resource here, any errors are unlikely. I'd recommend `catch(IOException ignored) {}`

app/src-gen/fr/gaulupeau/apps/Poche/entity/ArticleDao.java
```diff
@@ -125,14 +125,14 @@ protected void bindValues(SQLiteStatement stmt, Article entity) {
     /** @inheritdoc */
     @Override
     public Long readKey(Cursor cursor, int offset) {
-        return cursor.isNull(offset + 0) ? null : cursor.getLong(offset + 0);
+        return cursor.isNull(offset) ? null : cursor.getLong(offset);
```
di72nn: To correct minor issues in auto-genereted code (more importantly, in the code that is likely be auto-regerenated again) is to waste time.

tcitworld:
@Strubbl Please remove the `src-gen` files as they will be generated again anyway.

di72nn:
@Strubbl can you please rebase excluding `5c09abd`, `20f1a08` and `2800338` so not to clutter change history for `src-gen`?

Strubbl:
Yes, totally agree. I didn't see this in the inspect view. Why was src-gen checked in? 
I am going to rebase & then remove the folder.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
Strubbl: Okay, you are right. By now we do not care about this reader being closed. But we might care if we observe some strange bug. 
Does this log message have any bad side effects if we leave it in? I can also improve/reduce the message itself or change the log level, but i would not want to leave it out.

Strubbl:
The last 4 commits remove the content src-gen/ in the git repo. In future Git will ignore all files being generated in this dir (-->.gitignore). 
The directory has to be kept due to the greenDAO generator, which expects this src-gen/ to be existent. Otherwise we get an exception: `Exception in thread "main" java.io.IOException: ./app/src-gen does not exist. This check is to prevent accidental file generation into a wrong path.`.
Last commit adds the dependency to generate the DAO java files when creating the app.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
di72nn: In this case it doesn't really matter whether we have it logged or not. Even if the exception happens, I don't think the warning will tell us anything (we only care whether we could read the resource, not close it). It is a general advice, I don't mind having logging here.
I'd recommend `Log.w(TAG, "readRawString() failed to close reader", e);`. We pass the exception to logger and it handles the formatting for us.

di72nn:
Well, I don't mind having the generated `src-gen` in the repo. But if the call to `gen-dao` doesn't break incremental build (or whatever gradle do/will use), I'm fine with this.
The only (minor) issue I see now is that Android Studio shows `.keep-directory` in the source root. Maybe move it to `app/src-gen/fr/gaulupeau/apps/Poche/entity/`?

Strubbl:
From version control point of view I do not like to have generated code in the repository. That's why i wanted to remove that. 

The calls to gen-dao tasks work, at least in my machine and in travis CI. gen-dao tasks also run fine from command line or Android Studio. Do you mean something else or is that already okay for you?

I do not like that file .keep-directory being listed in the source root either. I did not recognize this so far. 
From tooling point of view, it is necessary to have the app/src-gen/ folder already created. I believe gen-dao does not care about the existence of any subfolders in that src-gen. So we could move the file to app/src-gen/fr/gaulupeau/apps/Poche/entity/. I am writing down a todo to check whether there is an option for Android Studio to ignore that file or if Android Studio already ignores specific place holder files. I will look into that more in detail tomorrow.

di72nn:
> The calls to gen-dao tasks work, at least in my machine and in travis CI. gen-dao tasks also run fine from command line or Android Studio. Do you mean something else or is that already okay for you?

I mean that gradle builds projects incrementally: it skips the parts that wasn't changed since the last build. Since the `gen-dao` is called for every build it may invalidate some built files, so gradle *might* do some excessive work (aside from calling the `gen-dao` every time). I did not notice any interference in the incremental build so far, but I did not thoroughly check it.

di72nn:
I fixed a `BatteryLife` warning: on Lollipop+ it will use JobScheduler to schedule a job (to run it when network becomes available), on older Android versions it will use the usual BroadcastReceiver. I **did not** test on Android 5+. It affects only "Separate auto-sync for local changes" option.

di72nn:
I removed one of my commits (in favor of #311) and force-pushed. Sorry for inconvenience (if any).

Strubbl:
I have this repo as upstream in my git and here's how i fixed my branch:

```
git checkout master
git branch -D fix-lint-issues
git checkout f24703f909bae5e2e7f5645540aa8db8e53c57c6
git checkout -b fix-lint-issues
git pull upstream fix-lint-issues
git push -u upstream fix-lint-issues
```

Strubbl:
Okay, i am closing this PR and think tackling lint issues in separate PRs is better. A long running PR is maybe not the best to track lint issues. 

There is no important work in here, which is going to be lost now, isn't it?

Issue #168, opened 2015-12-12
Eink friendly feature requests


gerroon:
Hello

The recent wallabag app is great and it is so much better than the previous version.

I am using it on my android eink device. It  is all good however these areas might help even further

- page down/up button support

- on screen font sizing option (they are undert a menu atm)

- bold font support 

- better and sharper font display for eink devices (maybe render them as black as possible?)

thanks

di72nn:
Hello,
Thank you for the suggestions.

> page down/up button support

Can you give some examples of the apps that support physical buttons (preferably open source apps)?

> better and sharper font display

Did you try "Light (high contrast)" theme? It uses black color for fonts.

[Continuation of #60].

gerroon:
@di72nn 

Hi

The page up/down buttons can be on screen or can utllize the volume buttons etc for paging. You can check Document Viewer for Android.

I am already using the contrast theme. It improves the look but I find it to be still a bit faint as far as the font rendering goes.

thanks

grote:
Hi, I also have an Android based E-Book Reader where I can install wallabag on. It would be awesome to be able to read stuff from wallabag on an E-Ink Screen.

Basically, the problem with E-Ink is that it does not do scrolling well. Refreshing the screen takes much more time than on other displays. That is why @gerroon asked for paging with buttons.

silberzwiebel:
I also have an android based e-reader ([inkbook prime](https://www.inkbook.eu/inkbook-prime)) and would like to read my wallabag articles with it.
The inkbook prime has four physical buttons (two on each side) that turn the page up or down. Using these buttons scrolls e.g. the app list in f-droid. I presume they send a code similar PageUp/Down on a PC keyboard? However, they do not work in the wallabag app. I could configure to scroll with the volume up/down buttons but this conflicts with other apps.
So, could you add support for this type of buttons, also? That would be great!

di72nn:
@silberzwiebel F-Droid app doesn't seem to use anything specific to handle keys (at least the new versions with fancy UI). Does your buttons do anything in wallabag at all? Can you use something like [KeyEvent Display](https://play.google.com/store/apps/details?id=aws.apps.keyeventdisplay) to gather some info? (Use proprietary apps at your own risk).

silberzwiebel:
> Do your buttons do anything in wallabag at all?

[...]

PR #299, opened 2016-09-20
Add tap-to-scroll option

di72nn:
#168.

Replaces #266.

Same note applies:

The "sides" are limited by:
 * height: (0.25; 0.75) scrollView's height;
 * width: left is [0; 0.3) of scrollView's width and right is (0.7; 1].

For some reason touch events are not consumed: if you tap on a link it will both scroll and open link menu.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
@@ -370,6 +370,22 @@ public class Settings {
         setString(R.string.pref_key_ui_theme, theme.toString());
     }
 
+    public boolean isTapToScrollEnabled() {
+        return getBoolean(R.string.pref_key_ui_tapToScroll_enabled, false);
+    }
+
+    public void setTapToScrollEnabled(boolean value) {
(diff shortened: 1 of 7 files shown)


di72nn:
Is there any problem with it?
I got the impression that the feature would be very useful for E-Ink users, but would not affect anyone who does not enable it. The implementation is somewhat poor (no UI buttons (are they needed?) and the event consumption issue (nothing serious)), but it is better than nothing.

P. S.: Now that I think about it, maybe I should also add "scroll with volume buttons" option or something like that.

Strubbl:
Sorry, i am not interested in this feature at all. That's why i did not review yet. 

The scroll with volume buttons option might be interesting for users without e-ink display, too.

tcitworld:
@di72nn This would be great !
> P. S.: Now that I think about it, maybe I should also add "scroll with volume buttons" option or something like that.

di72nn:
Added an option to scroll with volume buttons. Also added a "smooth scrolling" checkbox. Is it just me, or smooth scrolling with volume buttons is not always smooth (maybe threading issue or something else)?

PR #310, opened 2016-09-30
Settings: better application of runtime changes

di72nn:
Use onStart/Stop instead of onResume/Pause: the code was not called at the desired time.
Fix application of autosync enabled/interval changed: it was failing to correctly apply changes if only one of the options was changed.


Strubbl:
:+1: yep, good change. already with onstart the activity is visble

[lcm](https://developer.android.com/training/basics/activity-lifecycle/starting.html)

PR #311, opened 2016-09-30
Use JobScheduler on Android 5+

di72nn:
*I initially committed it to `fix-lint-issues` branch, but I think this change should be merged ASAP.*

I fixed a `BatteryLife` warning: on Lollipop+ it will use [`JobScheduler`](https://developer.android.com/topic/performance/background-optimization.html) to schedule a job (to run it when network becomes available), on older Android versions it will use the usual BroadcastReceiver. I **did not** test it on Android 5+. It affects only "Separate auto-sync for local changes" option.
Essentially, without `JobScheduler` the `CONNECTIVITY_CHANGE` event probably won't be delivered on Android 5+.


tcitworld:
I did not test it with Android 7 (I'm not sure the emulator can handle this kind of things anyway). Works properly on Android 6.

PR #282, opened 2016-06-27
[WIP] Decent offline functionality and other improvements

di72nn:
Ok, I finally started to implement it.

The first batch of changes:
* Update and add some dependencies: update greendao, okhttp, add greenrobot:eventbus.
* Update DB schema (add QueueItem), regenerate greendao classes.
* Add an IntentService and some basic logic to perform network-related operations in BG service and to enqueue failed actions for later sync.

I create this PR so we can discuss something about it in the future.
Currently I have some thing to implement, so code-related collaboration is not yet needed (not that I expect much).

TODO:
- [x] Use background service for the most operations.
- [x] Handle offline operations.
- [x] Basic auto-sync.
- [x] Decent notifications.
- [x] Basic PreferenceFragment-based settings.
- [x] Set `configurationIsOk` flag somewhere (requires configuration master/wizard?).
- [x] Unset `configurationIsOk` flag in Settings.
- [x] Delete obsolete translation lines.

Not so urgent/important TODO:
 - [Better temporary errors detection](#issuecomment-230098478).
 - Better/prettier/fancier notifications.
 - Do something with UI responsiveness (while background operation is performed).
 - Refactoring: reuse code.

Separate features/improvements:
 - Show last update time somewhere.
 - Differentiate Wi-Fi and data connection in auto-update?
 - Download PDF in BGService.
 - Use support library for Settings?

PR #318, opened 2016-10-06
Fix UI freezes

### LINKED ISSUE #315: UI freezes while BG operations are performed
I made 2 videos to show you what's happened on our smarthone for Wallabag when the network does a "yo-yo" (go away / come back)

On the second one, you will see "black" or "white" screen.
This is not due to the video encoding. It's the real effect I could notice when the smartphone looses is breath. To fix this behavior I finally have to remove Wallabag from the recent app I launched.
You can see the time elasped in the footer of the video, showing you the real time I have to wait before being able to do anything

https://foxmask.trigger-happy.eu/static/wallabag/2016_10_04_07_40_18.mp4
https://foxmask.trigger-happy.eu/static/wallabag/2016_10_04_07_41_58.mp4

di72nn:
I rewrote something I wanted to rewrite almost from the beginning of #282.

Action processing before this PR (does not apply to "sync queue" and "update feeds" actions):
 1. *(DB transaction is started here).*
 1. The offline queue is checked.
 1. If the action wasn't canceled out by queue contents, an attempt to transmit the action to server is performed.
 1. If the transmission fails, the action is added to the offline queue.
 1. *(DB transaction is ended here).*
 1. ...
 1. The queue is synced sometime later.

After this PR:
 1. *(This is done inside of one short DB transaction)* The offline queue is checked. If the action is not canceled out, it is added to the queue.
 1. If the queue is not empty, "sync offline queue" action is triggered.
 1. The queue is synced *(using only one short DB transaction to write changes at the end)*.

Essentially, the queue sync is the only way to pass local changes to the server. That change allowed to remove a lot of code duplication. And it also allowed to remove most of the DB transactions that presumably caused UI freezes. 
Though, the is a limitation: the is no (easy?) way to get the action result for a specific action. But it was barely used anyway.

Fixes #315.


app/src/main/java/fr/gaulupeau/apps/Poche/service/BGService.java
```diff
@@ -58,54 +58,40 @@ public BGService() {
...
-                } finally {
-                    removeStickyEvent(startEvent);
-                    if(result == null) result = new ActionResult(ActionResult.ErrorType.Unknown);
-                    postEvent(new SyncQueueFinishedEvent(actionRequest, result));
+            case Archive:
+            case Unarchive:
+            case Favorite:
+            case Unfavorite:
+            case Delete:
+            case AddLink:
```
tcitworld: There's no way to improve this switch condition ?

tcitworld:
Will test asap.

app/src/main/java/fr/gaulupeau/apps/Poche/service/BGService.java
di72nn: I considered moving these into the `default` case (another check is done inside of `serveSimpleRequest()` anyway), but I don't see improvements in such move. I thought that it will be easier to make a decision when there will be need to change this part again.

app/src/main/java/fr/gaulupeau/apps/Poche/events/SyncQueueFinishedEvent.java
```diff
@@ -6,6 +6,7 @@
 public class SyncQueueFinishedEvent extends BackgroundOperationEvent {
 
     protected ActionResult result;
+    protected Long queueLength;
```
di72nn: I use object types instead of primitives types when I need `null` to be a possible value (like "the value is not set"). Otherwise instead of `null` I would probably need to use some "special values" like `-1` which may not be always obvious to use in conditions. `value != null` is pretty much obvious no matter what.
When I'm sure the value is always set, I stick with primitive types.

app/src/main/java/fr/gaulupeau/apps/Poche/service/BGService.java
```diff
@@ -124,402 +110,232 @@ protected void onHandleIntent(Intent intent) {
...
-        } finally {
-            daoSession.getDatabase().endTransaction();
         }
 
-        if(queueChangedLength != null) {
-            postEvent(new OfflineQueueChangedEvent(queueChangedLength));
+        if(queueLength != null) {
+            postEvent(new OfflineQueueChangedEvent(queueLength));
+        } else {
+            queueLength = (long)queueItems.size();
```
di72nn: I cast `int` to `long` so the compiler could implicitly cast it to `Long`. You can't directly cast `int` to `Long`, though.

You can try `Long.valueOf(queueItems.size())`, but Android Studio will suggest to replace it with `(long)`. You can also use `(Long)(long)`, but that's just redundant.

app/src/main/java/fr/gaulupeau/apps/Poche/events/EventProcessor.java
```diff
@@ -202,6 +205,12 @@ public void onSyncQueueFinishedEvent(SyncQueueFinishedEvent event) {
 
         getNotificationManager().cancel(TAG, NOTIFICATION_ID_SYNC_QUEUE_ONGOING);
 
+        ActionResult result = event.getResult();
+        if(result != null && !result.isSuccess()
```
di72nn: Done.

app/src/main/java/fr/gaulupeau/apps/Poche/service/BGService.java
Strubbl: Thank you

Issue #321, opened 2016-10-13
wallabag does not differentiate between case of login name


Strubbl:
I just tried to login at @Zlitus wallabag instance with a user and pw i got from him. He sent me the user name in all small letters. 

Login works with that, but getting the rss feed throws a 404 error. Root cause for the 404 error is that the name in original is not all small letters, the first letter is a big letter.

Another side effect of mixing up the case-sensitivity in the login name is, that the connection assistant does not complete with a false case login. Only after fixing the login name, the connection assistant finishes with success, but after that see #320

di72nn:
I'm not sure I understand it clearly.
What I saw in gitter chat:
 * In the app username entered with a capital latter, let's say "**U**ser1".
 * The app populated feeds user id with the same value "**U**ser1".
 * The apps tries to access feed "https://host/**U**ser1/token/starred.xml", and fails because server returns `404` in for the URL with the capital letter in it.
 * In this particular case the user provided a URL of his feed (that he presumably copied from web UI) which had lower-case username "https://host/**u**ser1/token/starred.xml", the URL works fine.

I presumed it would be fixed by simple `feedsUserID = username.toLowerCase()` on the client.

I just tested feeds with wallabag 2.0.8 (I've not updated it yet) and it seems the server ignores username case (it ignores it on login as well as for feeds URL). Was it changed in the recent versions?

I've just tried to register on v2.wallabag.org, but for some reason I can't either login or finish the registration (does it requires e-mail confirmation?) - I don't get anything on e-mail even for password reset request.

I can't reproduce the behavior and I'm not sure I understand what you described. If the server accepts login in any case to log user in, but not for feeds, then the server should be consistent and accept feeds URL with the username in any case, because the app shouldn't guess it.

Strubbl:
> I just tested feeds with wallabag 2.0.8 (I've not updated it yet) and it seems the server ignores username case (it ignores it on login as well as for feeds URL). Was it changed in the recent versions?

I would say yes, because of the logging from my test case. Zlitus wallabag version was the most recent, so it was 2.1.1
maybe @nicosomb @j0k3r or @tcitworld can answer this? 

> I can't reproduce the behavior and I'm not sure I understand what you described. If the server accepts login in any case to log user in, but not for feeds, then the server should be consistent and accept feeds URL with the username in any case, because the app shouldn't guess it.

[...]

PR #323, opened 2016-10-13
Trim Feeds UserID instead of blindly replacing it in getCredentials()

di72nn:
Use somewhat more elaborate hack to get Feeds UserID.
It still won't work if username is empty (only HTTP Auth used).
#321.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagServiceEndpointV2.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagServiceEndpointV2.java
@@ -102,25 +102,42 @@ public class WallabagServiceEndpointV2 extends WallabagServiceEndpoint {
         FeedsCredentials fc = getCredentials("/config", CREDENTIALS_PATTERN);
         // fc.userID might include the subdirectory in which wallabag is installed.
         // If feed url is "https://example.com/wallabag/complex/user/name/token/unread.xml",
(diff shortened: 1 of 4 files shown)


di72nn:
Now it attempts to use `httpAuthUsername` if no `username` is available.

app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagServiceEndpointV2.java
```diff
@@ -100,15 +100,44 @@ public ConnectionTestResult testConnection()
...
     public FeedsCredentials getCredentials() throws RequestException, IOException {
         FeedsCredentials fc = getCredentials("/config", CREDENTIALS_PATTERN);
-        // overwrite userID with username because first matcher group of previous regex, which
-        // should return the user name, might include the subdirectory in which wallabag is installed
-        if(fc != null) fc.userID = username;
+        // fc.userID might include the subdirectory in which wallabag is installed.
+        // If feed url is "https://example.com/wallabag/complex/user/name/token/unread.xml",
+        // fc.userID will be "wallabag/complex/user/name", but should be "complex/user/name".
+        // The following code is an attempt to work around this issue.
+        if(fc != null) {
```
Strubbl: since we have no else block here, i suggest inverting the condition and directly return if fc is null. This way we can move the whole content of this if block some spaces left and reduce the nesting.

di72nn:
Good point, done.

PR #328, opened 2016-10-15
Update to greenDAO 3

Strubbl:
Please review

@di72nn can you please have a closer look to the BGService. Is it fatal if we use non-exclusive transactions now?

#edit new PR because i pushed to this repo in order to get your help in this issue.


app/src/main/java/fr/gaulupeau/apps/Poche/ui/IconUnreadWidget.java
```diff
@@ -27,8 +27,9 @@ static void updateAppWidget(Context context, AppWidgetManager appWidgetManager,
                                 int appWidgetId) {
         Log.d(TAG, "updateAppWidget() appWidgetId=" + appWidgetId);
 
-        long unreadCount = DatabaseUtils.queryNumEntries(
-                DbConnection.getSession().getDatabase(), "ARTICLE", "archive=0");
+        long unreadCount = DbConnection.getSession().getArticleDao().queryBuilder().
+                where(ArticleDao.Properties.Archive.eq(0)).count();
```
di72nn: We probably should use `false` instead of `0` here.

di72nn:
`beginTransaction()` should be fine - I made the transactions as short as I could the last time I changed it. Also non-exclusive transactions might have been exclusive anyway, I'm not exactly sure about it :/ (it is unlikely to have any noticeable affect for now, so I don't bother checking it yet).

di72nn:
BTW, I forgot to delete the generated `OfflineURL` and `OfflineURLDao` classes.

di72nn:
I think we should release a new version (beta/whatever, I just want a signed APK and an F-Droid build if possible) before merging it, so we have a version to test with, if something with greendao 3 goes awry. We as developers can test with whatever we build, but if we would want to ask a user to reproduce some potential bug with pre-greendao3 version, we would need an APK (or specific version name for F-Droid users).

Strubbl:
I saw we have a new tag 1.11.0, which should fulfill @di72nn recommendation to "release a new version [...] before merging it". Apart from solving the conflict in  build.gradle, we can now merge this PR, can't we?

Strubbl:
I used the option of checking it out via command line to resolve conflicts and perform a manual merge.

PR #331, opened 2016-10-15
Replace greendao generator with gradle plugin

di72nn:
Should we also move `entity` package into `data`?


Strubbl:
Good question. On the one hand side yes, because queue item and article hold data like the other classes in data/. On the other hand side those two are the only ones being "real" entities being saved in the db. Hmm.

di72nn:
I put everything DAO-related into separate packages, but "inside" `data`.
Entities inside `data.dao.entities` are shown in the main source tree and are distinct from other classes in `data`.
Generated classes inside `data.dao` are shown by AndroidStudio in a separate source tree (since these are generated sources in the `build` directory).

Main sources tree:
 * `data.dao.entities`
   * `Article`
   * `QueueItem`

Generated sources tree:
 * `data.dao`
   * `DaoMaster`
   * `DaoSession`
   * `ArticleDao`
   * `QueueItemDao`

Issue #303, opened 2016-09-24
Show last update time


di72nn:
Show somewhere the time of last (auto-)update.

PR #335, opened 2016-10-23
Show last update time #303

GoncaloGarcia:
I added a text field to the ArticleList View to indicate when the last update happened.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
@@ -551,6 +551,14 @@ public class Settings {
         setBoolean(R.string.pref_key_internal_offlineQueue_pending, value);
     }
 
+    public long getLastFeedUpdateTimestamp() {
+        return getLong(R.string.pref_key_internal_lastFeedUpdateTime, -1);
+    }
+
+    public void setLastFeedUpdateTimestamp(long value) {
+        setLong(R.string.pref_key_internal_lastFeedUpdateTime, value);
+    }
+
     public boolean isHandlingHttpScheme() {
         return context.getPackageManager()
                 .getComponentEnabledSetting(getHttpSchemeHandlingComponent())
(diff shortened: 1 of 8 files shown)


di72nn:
[Created a PR](https://github.com/GGfpc/android-app/pull/1). If it gets merged, there would still be things to fix:
 * Time probably should be in the format like "5 minutes ago"/"1 day ago"/etc..
 * Maybe we can move the UI element somewhere else? Currently it occupies to much space (horizontal-wise).

GGfpc:
I wanted to push to the branch you created, not sure why it ended up here. Anyway, I fxed the relative time span and also changed the position of the TextView. Since you said it used up too much space I added a fade out transition after 2 seconds.

di72nn:
Sorry for my late reply.

I personally not sure whether the text should be hidden. Now that the update text may be longer than just time, the screen space usage is somewhat more justified.
Also, when the TextView is placed under TabLayout, part of the pull-to-refresh animation is getting stuck on the TextView (if the pull-to-refresh is cancelled; probably lack of redraw or something).
Maybe we can place the TextView above TabLayout but make it look like part of title bar (or actually make custom title bar with the text included)? @tcitworld @Strubbl any opinion on that?


> I wanted to push to the branch you created, not sure why it ended up here.

That's how it should be. Whatever you push to your branches shows up in the corresponding PRs (if any) - that's how github works. And you can't push to my branches.

Strubbl:
I do not know, if the article list view is the best view for this piece of information. I would give the maximum space to the list view to see as many articles from the list as possible. 

Maybe if we do not enlarge the title bar, but make the title "wallabag" a little smaller and at the info about last sync time as subtitle in small letters. 

@GGfpc Do you have any suggestions for placing the last update time into the view? Can you add some screenshots for different suggestions?
Or maybe we can add that time to the menu as greyed out item?

GGfpc:
I believe the best thing to do would be to add it to the title bar. Its unobtrusive and that way it could be permanent. I'm not sure how that works on Android though.

tcitworld:
Will have a look on how it looks.

Strubbl:
Currently there is a fade out animation after the start of the application. This looks cool it think.
But the issue is that it is only shown after an update or when onCreate of that view is called. Because duing daily usage my app does not restarted, I suggest adding this animation to the onResume activity, too.

ngosang:
I prefer to display that info in the sidebar to save space in article list.

Strubbl:
I guess this issue can be closed.

di72nn:
@GoncaloGarcia thanks for your work anyway.

PR #348, opened 2016-11-13
small updates and fixes for Spanish translation

auroszx:
This is just a small update translating the new strings and fixing some errors I didn't notice before. Please let me know if anything should be fixed. Thank you.

### DIFF
--- a/app/src/main/res/values-es/strings.xml
+++ b/app/src/main/res/values-es/strings.xml
@@ -2,6 +2,12 @@
 <resources>
     <string name="app_name">wallabag</string>
     <string name="label_name">¡Guárdalo!</string>
+<!--
+    <string name="author">
+GAULUPEAU Jonathan — 2013-2015
+\nVictor Häggqvist — 2015
+</string>
+-->
     <string name="btnMarkRead">Marcar como leído</string>
     <string name="btnMarkUnread">Marcar como no leído</string>
     <string name="btnSettings">Configuración</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-es/strings.xml
```diff
@@ -2,6 +2,12 @@
 <resources>
     <string name="app_name">wallabag</string>
     <string name="label_name">¡Guárdalo!</string>
+<!--
```
tcitworld: Please remove this.

app/src/main/res/values-es/strings.xml
```diff
@@ -211,8 +222,10 @@
 
     <string name="connectionWizard_providerSelection_titleMessage">Selecciona un proveedor de wallabag</string>
     <string name="connectionWizard_providerSelection_provider_general_name">Otro</string>
-    <string name="connectionWizard_providerSelection_provider_general_desc">Selecciona esta opción para introducir detalles manualmente</string>
+    <string name="connectionWizard_providerSelection_provider_v2wallabagorg_name" translatable="false">v2.wallabag.org</string>
+    <string name="connectionWizard_providerSelection_provider_general_desc">Selecciona esta opción para introducir los ajustes requeridos manualmente</string>
     <string name="connectionWizard_providerSelection_provider_v2wallabagorg_desc">Selecciona esta opción si tienes una cuenta en https://v2.wallabag.org</string>
+    <string name="connectionWizard_providerSelection_provider_framabag_name" translatable="false">Framabag.org</string>
```
Strubbl: Does anybody know if we really need the strings which are `translatable="false"` in every language file? I would remove them, too. But i do not know if this has any impact if those strings are only available in the values/strings.xml

di72nn: `translatable="false"` strings should be present only in default locale files (in `res/values/` directory).

Strubbl:
Thank you very much for your contribution.

auroszx:
No problem. I'll do what I can to keep the translation up to date.

PR #365, opened 2016-12-04
Replaced v2.wallabag with wallabag.it

nicosomb:
v2.wallabag.org will be closed in next days. We need to replace the URL in the android application.

PR #367, opened 2016-12-07
Replace all v2.wallabag.org mentions

di72nn:
A follow-up to #365.

Also replace "app.wallabag.it" with "wallabag.it" in display-only strings. Users don't care about the "app." part, do they?

cc @nicosomb


nicosomb:
If they don't need to type the URL, OK for that.

PR #369, opened 2016-12-09
Start QR code scanner from app

di72nn:
In a feat of procrastination I wrote some code for the app to invoke QR-code scanner.
I added a button on one of the wizard pages. Suggestions are welcome.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/preferences/ConnectionWizardActivity.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/preferences/ConnectionWizardActivity.java
@@ -1,8 +1,10 @@
 package fr.gaulupeau.apps.Poche.ui.preferences;
 
 import android.app.Activity;
+import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
+import android.net.Uri;
 import android.os.Bundle;
 import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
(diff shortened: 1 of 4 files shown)


app/src/main/java/fr/gaulupeau/apps/Poche/ui/preferences/ConnectionWizardActivity.java
```diff
@@ -132,6 +169,25 @@ private ConnectionData parseLoginData(String connectionUri) {
         return new ConnectionData(values[0], values[1]);
     }
 
+    public void scanQrCode() {
+        try {
+            Intent intent = new Intent("com.google.zxing.client.android.SCAN");
```
tcitworld: Does that enforce a specific QR scanner client ?

tcitworld:
Needs rebase to have only the second commit.

di72nn:
It was a PITA to set up the app multiple times during development, so I updated this PR. I also added support for password. With these changes one can set up his account in one swift press-the-button-find-the-qr-code-point-the-phone-at-the-code motion.

The intent works with **multiple QR code scanning applications**: a list of apps is presented by Android. I can see "Barcode Scanner" and "Binary Eye" (I don't think I have any other installed).

di72nn:
Added `intent.removeExtra(EXTRA_FILL_OUT_FROM_SETTINGS);` which I missed yesterday. Without it the feature didn't work if the wizard wasn't started from settings.

PR #372, opened 2016-12-17
Update translations

di72nn:
Thanks to @naofum for Japanese translation.

*I only checked whether the file format is correct, I did not check the translations' text.*


di72nn:
Updated & force-pushed.

Issue #375, opened 2016-12-19
Don't retry to delete article if 404 received -> high battery drain


shtrom:
I am seeing an odd behaviour due (probably) to deleting an article both in the Android app, and the Web interface before it synced.

Now, the Android app repeatedly tries to delete the article, as seen in my webserver logs:

    wallabag.example.net 192.0.2.1 - - [19/Dec/2016:06:25:56 +0100] "GET /delete/3992 HTTP/1.1" 200 818 "-" "okhttp/3.4.1"
    wallabag.example.net 192.0.2.1 - - [19/Dec/2016:06:26:01 +0100] "GET /delete/3992 HTTP/1.1" 200 818 "-" "okhttp/3.4.1"
    wallabag.example.net 192.0.2.1 - - [19/Dec/2016:06:26:07 +0100] "GET /delete/3992 HTTP/1.1" 200 818 "-" "okhttp/3.4.1"

Oddly enough, the application sees this as 404 error, as per the logcat (to follow), and keeps popping an unknown error (also to follow) before retrying to sync immediately. This leads to a high battery drain.

There are several things that could be done to address the issue.
* HTTP DELETE should be idempotent, and not return a 404, or any indication that the object to delete was not there, to the client. This is a GET happening here, but it should probably follow the same semantics
* In the current situation, if the Android app receives a 200 from a `GET /delete/...` wrapping a 404, this means that Wallabag has responded properly, but the item is no longer there, there should therefore be no need to retry the delete.
* Add some sort of incremental back-off in case of repeated sync failures to avoid draining the battery.
* Add a way to clear the app's sync queue, so one can manually cancel local sync actions that might be causing this sort of issue.

shtrom:
```
I/ActivityManager( 2537): Start proc 5225:fr.gaulupeau.apps.InThePoche/u0a66 for service fr.gaulupeau.apps.InThePoche/fr.gaulupeau.apps.Poche.service.WallabagJobService
D/WallabagServiceEndpointV2( 5225): executeRequestForResult() got response
W/WallabagServiceEndpoint( 5225): checkResponse() response is not OK; response code: 404, response message: Not Found
W/BGService( 5225):     at fr.gaulupeau.apps.Poche.network.WallabagServiceEndpoint.checkResponse(WallabagServiceEndpoint.java:124)
W/BGService( 5225):     at fr.gaulupeau.apps.Poche.network.WallabagServiceEndpoint.checkResponse(WallabagServiceEndpoint.java:115)
[...]

PR #376, opened 2016-12-19
Don't stuck on deleted articles

### LINKED ISSUE #319: Actions for articles removed on server are stuck in offline queue
1. Remove an article on server.
2. Try to remove / favorite / mark as read it in the app.

Expected result: no errors, article is changed offline until "full sync" is done (probably).

Actual result:
"Unknown error" is shown. The offline queue sync is blocked (no future changes will be synced until the erroneous action is somehow removed).

Cause: server (v2) responds with HTTP status code `404` for any action with removed articles. (I did not test it against v1 server).

This is pretty serious, because there are no way for user to get rid of it except to clear the app data (at least I'm not aware of any other way).

di72nn:
If article related actions (mark as favorite/archived, delete) fail because of HTTP 404, consider the action as successful.

v1.9 endpoint seems to be unaffected by the issue.

Fixes #319 #375.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagServiceEndpoint.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagServiceEndpoint.java
@@ -2,11 +2,10 @@ package fr.gaulupeau.apps.Poche.network;
 
 import android.util.Log;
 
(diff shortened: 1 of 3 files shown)


tcitworld:
I don't see any other way. 👍

PR #380, opened 2016-12-19
double default limit of article list items

Strubbl:
This is a a setting i am adjusting every time i install the app freshly. Do you agree on increasing the default limit of article list items?

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
@@ -351,7 +351,7 @@ public class Settings {
     }
 
     public int getArticlesListLimit() {
-        return getInt(R.string.pref_key_ui_lists_limit, 50);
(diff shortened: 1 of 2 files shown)


app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
```diff
@@ -351,7 +351,7 @@ public void setArticleFontSerif(boolean value) {
     }
 
     public int getArticlesListLimit() {
-        return getInt(R.string.pref_key_ui_lists_limit, 50);
+        return getInt(R.string.pref_key_ui_lists_limit, 100);
```
di72nn: This is unlikely to affect anything, not sure why I even bothered to set it. The actual default value is here: The actual default value is https://github.com/wallabag/android-app/blob/master/app/src/main/res/xml/preferences.xml#L96.

di72nn:
I don't have anything against changing this default value.

PR #381, opened 2016-12-19
change default theme to dark

Strubbl:
My suggestion is to change the default theme to dark. I know two friends using the app in dark theme and i am using the dark theme, too. But i do not know if this is common usage. It looks much better and is not too bright for the eyes. Therefore i would choose to change the default.

What do you think?

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
@@ -368,7 +368,7 @@ public class Settings {
             } catch(IllegalArgumentException ignored) {}
         }
 
-        return theme != null ? theme : Themes.Theme.Light;
+        return theme != null ? theme : Themes.Theme.Dark;
     }
 
     public void setTheme(Themes.Theme theme) {
--- a/app/src/main/res/xml/preferences.xml
+++ b/app/src/main/res/xml/preferences.xml
@@ -78,7 +78,7 @@
             android:key="@string/pref_key_ui_theme"
             android:title="@string/pref_name_ui_theme"
             android:dialogTitle="@string/pref_name_ui_theme"
-            android:defaultValue="Light" />
+            android:defaultValue="Dark" />
         <fr.gaulupeau.apps.Poche.ui.preferences.IntEditTextPreference
             android:key="@string/pref_key_ui_article_fontSize"
             android:title="@string/pref_name_ui_article_fontSize"


app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
```diff
@@ -368,7 +368,7 @@ public void setArticlesListLimit(int limit) {
             } catch(IllegalArgumentException ignored) {}
         }
 
-        return theme != null ? theme : Themes.Theme.Light;
+        return theme != null ? theme : Themes.Theme.Dark;
```
di72nn: Same as https://github.com/wallabag/android-app/pull/380#discussion_r93113848: https://github.com/wallabag/android-app/blob/master/app/src/main/res/xml/preferences.xml#L81

di72nn:
I have doubts about it. Is there any guideline about preferred default themes?

Strubbl:
Yes, kind of. If you use material theme, the default is dark and you have to explicitely choose ligth if you want, see also: https://developer.android.com/training/material/theme.html

> The material theme is defined as:
> 
>     @android:style/Theme.Material (dark version)
>     @android:style/Theme.Material.Light (light version)

tcitworld:
Not sure about this... Most apps I know are have the default light theme enabled. It's just a category of (power ?) users that choose the dark theme.

Lokesh-Krishna:
Apps using the dark theme by default are definitely the exception, not the norm. Also, the option to switch already exists so is there really any need to make it the default?

Strubbl:
> Not sure about this... Most apps I know are have the default light theme enabled. It's just a category of (power ?) users that choose the dark theme.

> Apps using the dark theme by default are definitely the exception, not the norm. 

Do you have proof for your assumption? I have not. And of course not for the opposite setting.

What kind of apps are you considering? I mean you cannot compare games or apps where you are not reading with an app where reading is the main purpose.

> Also, the option to switch already exists so is there really any need to make it the default?

It is considered because i would like to give the standard user a reasonable default. Having the choice is only nice for power users as tcitworld mentioned. But standard users do not care and want the app just to work. In my opinion dark theme is better for reading. That is why i started this discussion with this pull request. 

So far all participants in this issue have no valid data or reference on which might be the default for more users. Maybe there is some Android design guru having some recommendation reference/source. The only source is now: https://developer.android.com/training/material/theme.html

>    The material theme is defined as:
>
>    @android:style/Theme.Material (dark version)
>    @android:style/Theme.Material.Light (light version)

I quoted this, because the Android people could have made the opposite decision and make Theme.Material (light version) and Theme.Material.Dark (dark version). But they did not.

In the end, it is only a proposal and can be closed without merging if it makes no sense.

Lokesh-Krishna:
> Do you have proof for your assumption? 
> I have not. And of course not for the opposite setting.

Well, I've used Pocket for a long time and the light theme was the default there. The same is the case with many other apps whose sole purpose is reading, like Instapaper, Play Books, Kindle, Moon+, Aldiko, etc.

I guess you could also use the web itself as an example because most content oriented websites default to a black text on white background with most not even offering a dark theme.

nicosomb:
I read my articles with dark theme. 

But I think that the default value should be light theme (twitter applications like Twitter or Talon are in light mode, etc..

tcitworld:
Okay I can't decide for myself so I'm going to ask for @nicosomb to publish a poll on Twitter.

tcitworld:
ping @nicosomb 
also please rebase @Strubbl

nicosomb:
https://twitter.com/wallabagapp/status/819172730591412225

tcitworld:
People have spoken. Please rebase once more and merge.

di72nn:
I don't care about default theme, but for the record: the vote is almost even.

tcitworld:
Yeah, I'm starting to wonder if we shouldn't in fact add the choice during the connexion assistant.

Strubbl:
I suggest changing nothing in case the vote is almost even. If none of the options can get significant more votes than the other, there's no need to change.

To keep setup process as easy as possible, i would not add it to the assistant.

Strubbl:
54 votes and 50/50. So i am closing this PR now.

PR #384, opened 2016-12-19
add source app info to notifications

Strubbl:
This commit adds the information to a notification that wallabag is the
source of that notification. Otherwise notification from similar apps
cannot be distinguished.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/events/EventProcessor.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/events/EventProcessor.java
@@ -152,11 +152,11 @@ public class EventProcessor {
         Context context = getContext();
 
         FeedUpdater.FeedType feedType = event.getFeedType();
-        String detailedMessage;
+        String detailedMessage = context.getString(R.string.app_name) + " ";
         if(feedType == null) {
-            detailedMessage = context.getString(R.string.notification_updatingAllFeeds);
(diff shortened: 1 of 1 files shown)


di72nn:
Maybe add a colon after the app name?

Strubbl:
> Maybe add a colon after the app name?

Yes i would like to, but the french people add a space before the colon, german ones usually do not. That's why i left it out. 
any idea for a general divider?

//cc @nicosomb @tcitworld

tcitworld:
Doesn't matter much. Anyway, on Android 7.1 the notifications show up from which app they come from :

di72nn:
> any idea for a general divider?

Hyphen / minus sign / dash?

> Then we should prepend app name only if Android version <= 6 (?).

on Android 7.1 the notifications show up from which app they come from

tcitworld:
@Strubbl Please add an hyphen. :)

Strubbl:
Yes, if i am back to my android development PC i am going to do that. Currently i have only a notebook with me. And i think a middle dot (this one · ) is going to be better than a hyphen. And i also need to add the android api version IF

tcitworld:
Ping @Strubbl

Strubbl:
changes implemented. I have used the middle dot and i hope this is okay? It is smaller compared to a hyphen or the like. Also the wallabag app name is only added when using Android version before Nougat.

tcitworld:
Restarting the build, travis had an issue.

Issue #317, opened 2016-10-05
Update app icon and support the round icon of Android 7.1


jlnostr:
Besides the current material icon request in the app (#272 #273), I'd like to see an updated app icon as well. In my opinion a white "w" on a black background should be enough to recognize wallabag. And with Android 7.1 in sight, it would be useful to create an round icon as well :)

Lokesh-Krishna:
What are the specifications an icon should meet for Android? I believe any number of icon pack developers for GNU/Linux wouldn't mind making an icon as long as they're provided with the exact specifications.
I've been waiting long time to see a Material Design app icon.

jlnostr:
I think this is the right documentation for it: https://material.google.com/style/icons.html#icons-system-icons

Lokesh-Krishna:
Thank you so much!
I'll contact an icon pack developer to see what can be done :)

Lokesh-Krishna:
Thank you so much!
I'll contact an icon pack developer to see what can be done :)

tcitworld:
Also, maybe [icon shortcuts](https://developer.android.com/preview/shortcuts.html).

Lokesh-Krishna:
@tcitworld I don't think that's something an icon designer should be doing. A separate issue maybe?

varlesh:
![image](https://cloud.githubusercontent.com/assets/8083855/21073340/cfa25baa-bef4-11e6-8564-06cce9e36a2f.png)

[Wallabag.zip](https://github.com/wallabag/android-app/files/643744/Wallabag.zip)

tcitworld:
I really like the general aspect, but could the wallaby be just a little less large ? A thousand thanks for your work.

varlesh:
you mean large "W"?
Soryy my English very bad ;)

tcitworld:
Apologies. I had found the kangaroo itself a bit large at first on your picture, but in fact it's because of the way it's stylized on our own logo with the "w" :
![](http://www.loeuillet.org/images/stickers.jpg)

tcitworld:
Suits me well on second look. :)

jlnostr:
And what do you guys think about the plain "w" (with a long shadow) on a black background?

varlesh:
Ok. I use #2780e3 color for circle background - it's your main color on site.

varlesh:
Black not good for material... IMHO

jlnostr:
[...]

PR #392, opened 2016-12-26
[WIP] Updated color scheme

Lokesh-Krishna:
This is with reference to the discussion at #317.

@di72nn I finally managed to figure it out and I couldn't have without you so thank you so much!

I edited the theme using the Theme Editor in Android Studio and this applied well on both, the Light and the Dark theme.

Is this an acceptable method of doing this or is this there a preferred method?

### DIFF
--- a/app/src/main/res/values/values.xml
+++ b/app/src/main/res/values/values.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="primary_material_dark">#2196f3</color>
+    <color name="accent_material_dark">#2962ff</color>
+    <color name="primary_dark_material_dark">#1976d2</color>
+    <color name="accent_material_light">#2962ff</color>
+</resources>


di72nn:
My experience in Android UI/theming is very limited, so...

The colors in the "Light" theme seem fine.

The "Dark" theme have the same blue action bar (in the "Dark (high contrast)" theme it's black). So we probably need to explicitly set the color for the "Dark" theme.

Also we have [colors.xml](https://github.com/wallabag/android-app/blob/master/app/src/main/res/values/colors.xml), so we can put the values there.
And the Android Studio suggests to add `tools:ignore="PrivateResource"` because the values override something.

Lokesh-Krishna:
I too got that message about adding `tools:ignore="Private Resources"` but I didn't understand where it should be added.

Who here has the most experience when it comes to UI? This is my first attempt at this so I could really use some advice. Will adding this to the colors.xml file have the same effect as this values.xml file?

di72nn:
We should probably set colors the way other themes do it (https://github.com/wallabag/android-app/blob/master/app/src/main/res/values/styles.xml).
There are various guides ([example](https://developer.android.com/training/material/theme.html#ColorPalette)). I have no time to get into it right now.

Lokesh-Krishna:
I'll go through the guide and try to make a pull request that is more in line with the other themes. Should I close this one then?

di72nn:
I'm fine with this PR being open until a new one is created.

Lokesh-Krishna:
I no longer have access to a system that can run Android Studio so would anybody else be interested in getting this done in the right manner?
I think I have picked the right colour so if it can only be implemented in a manner that fits in with the way that themes are currently specified then we should finally be good to go for the new icon.

PR #343, opened 2016-11-04
[wip] Image caching

Strubbl:
okay, i just implemented a stupid image downloader which has no queue and separate no thread yet. 
This is a TODO. I just want to save work here and maybe get some feedback or (even better would be) support from somebody (looking at @di72nn or @tcitworld :wink: )

I am not working on this branch again before sunday evening or even monday, if somebody wants to contribute some code.

PR #394, opened 2016-12-28
Background improvements and image caching

di72nn:
* Added secondary background service (purpose: a service that does stuff that should not block important actions like "add to queue").
 * Moved "download PDF" functionality to the secondary service.
 * Rebased image caching (#343) on top of it (I squashed it to avoid changes all over the place).


tcitworld:
> secondary background service

👍

di72nn:
Things need to be done:
 * Test recent changes.

Things may be done later:
 * Better external storage path detection / customizable path.
 * Housekeeping.
 * More refactoring.

di72nn:
Closed in favor of #396.

Issue #76, opened 2015-02-16
Locally download/cache pictures


tcitworld:
They aren't displayed when offline.

di72nn:
Done.

PR #396, opened 2017-01-07
Development

di72nn:
Incorporates #394.

Also includes some minor changes.

#76.


tcitworld:
Looks good to me.

Could the downloading pictures notification have a progress from 0 to the number of articles on which to fetch pictures ? Of course, some articles will have many pictures and some won't have any, but that would be a relatively good indicator of how long the download takes.

di72nn:
Done.

PR #399, opened 2017-01-12
Update transtations

di72nn:


tcitworld:
Strange that the German translation has one string less than Japanese or Russian.

di72nn:
French, German and Spanish are not translated completely.

tcitworld:
So is that an issue with Weblate committing `Currently translated at 100.0% (206 of 206 strings)` whereas there should be 207 ?

di72nn:
I presume the reason is that the commit is old.

tcitworld:
Right.

Issue #127, opened 2015-11-15
[1.9] Combine URL upload and sync


khinsen:
I don't see any good reason to have the upload of locally stored URLs as a separate operation that must be launched manually. I'd expect this to be part of the sync operation. It should first upload the locally stored URLs, and then download articles via the RSS stream.

Moreover, I'd appreciate an option to sync automatically when the app is started.

tcitworld:
Yup, it's just at first all processes were separated. Will change that.

tcitworld:
Related to #7.

di72nn:
Implemented it 1.12.0.

> Moreover, I'd appreciate an option to sync automatically when the app is started.

Track #360 and #404 for that.

Issue #113, opened 2015-11-12
Search


tcitworld:
Just search, but with filters however.

Issue #42, opened 2014-10-05
Sorting modes of article list


aisbergde:
- [x] by creation date.
- [ ] by update date.
- [ ] by estimated reading time.
- [ ] by name (alphabetical).

nr458h:
I also would like to add the ability to sort the archiv by marked as read time

Strubbl:
Currently this is not possible, because marked as read time is only provided starting with wallabag v2.4.0 (see https://github.com/wallabag/wallabag/pull/3630 )

So having wallabag server 2.4 release is a dependency to this issue - besides that the API-wrapper-lib needs to be adjusted for that purpose, too.

techexo:
Estimated reading time would be great, I think this option is long overdue (milestone 1.x! 😀).

Strubbl:
Milestone updated. :smile: 
Maybe it's easier to implement this after @di72nn has been using android paging lib mentioned in https://github.com/wallabag/android-app/pull/1014#issuecomment-712418757

Issue #17, opened 2014-05-10
Sorting mode of the articles


acotte:
I really don't understand how articles are sort. If we had to resort to only one sorting mode. Date of add in Wallabag is the best one for me.  More options is a plus.

tcitworld:
More sorting ways : estimated reading time, alphabetical.

Strubbl:
issue #42 also deals with sorting, but by name. Creation and update date should be possible, too.

di72nn:
Closed in favor of #42.

PR #403, opened 2017-01-19
Update lists

di72nn:
Adds "endless scrolling" instead of limited list. Memory may still be an issue.

Use `DiffUtil` to update lists partially.

I'm not completely sure about the implementation, so suggestions are welcome.

~~Also, I think I had an issue with the lists not being updated after actions like "open an article - add it to favorite - go back to lists". I couldn't reproduce it yet. I added logging of warning messages "fragment is null". I'm pretty sure it should either update list or log the warning. I have no idea why the fragments can be null after they were initialized (initially I thought, that it might be related to threading, but only the main thread is supposed to access fragments).~~ Probably fixed with 2630998.

Related issues:
* Basic search: partly #113.
* Sync queue before full update: #127.
* Sorting order: partly #17 and #42.

Testing needed.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -22,6 +22,8 @@
 
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <meta-data android:name="android.app.searchable"
+                       android:resource="@xml/searchable" />
         </activity>
         <activity
             android:name="fr.gaulupeau.apps.Poche.ui.ReadArticleActivity"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -19,6 +19,7 @@
         <activity android:name="fr.gaulupeau.apps.Poche.ui.ArticlesListActivity">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
+                <action android:name="android.intent.action.SEARCH" />
 
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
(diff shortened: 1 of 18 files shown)


di72nn:
Added basic search (using SQL's `LIKE %search string%`) on article title and text.
Current list is updated on every search string change - not sure about performance.

tcitworld:
Hmm. I guess it will provoke some performance issues with huge articles, some maybe limit to title for now. Anyway we'll probably propose a search API soon on the server so it will have to be integrated with it at some point.

di72nn:
Added an option to reverse sort order. The UI part should probably be improved (as for now it is just a menu item that reads "Change sort order"). Suggestions welcome.

There is an issue that fragments reinit their lists after orientation change. It is not critical (slight delay), but I'll have to fix it.

di72nn:
Started to use AsyncTask to update lists. Synchronization issues are possible.

Strubbl:
is the read to review label still valid atm?

di72nn:
I removed the async loading from this PR - now it is ready for review.

tcitworld:
Looks good. Will review and eventually build tomorrow.

di72nn:
@tcitworld 
> Hmm. I guess it will provoke some performance issues with huge articles, some maybe limit to title for now.

I forgot to answer that. It seemed to me that the performance was decent even on old devices, so I left it as is. I considered "powerful" search (being able to search in content) more important than minor performance issues (which don't affect anything but search). Though, anyone is free to improve the feature.

tcitworld:
I tested and it was okay for me so I merged. :)

Issue #360, opened 2016-11-30
Automatic update when the app is opened.


di72nn:
// AFAIR it was already mentioned not long ago, but I haven't found a separate issue.

 - [x] Auto-update.
 - [ ] On unmetered connections only (Wi-Fi, etc.).

di72nn:
#346.

PR #370, opened 2016-12-14
v2 API

di72nn:
I started to work on [v2 API wrapper library](https://github.com/di72nn/wallabag-api-wrapper) a couple of days ago. The library's API is far from being finished.

Here's a proof of concept. The android app that performs "Favorite/Unfavorite" via API (I've done basic implementation for the most of the v2 API methods in the library, but not yet integrated it in the app).

Changes:
 - Moved to v2 API. Dropped v1.9 support.
 - Rewritten update mechanism: faster (supposedly), more consistent (no updates for separate feeds that screw things up) updates.
 - Update on startup (#360).
 - Basic tags support (list by tag, add/modify/delete tags on articles) (#116).
 - Changed main activity UI (in order to have better navigation between tags and other lists).
 - Notifications updated a bit (progress added, etc.).
 - Other minor changes.

New opportunities (**not yet implemented** features that are now possible):
 - Show tags in lists and in article view.
 - Change article title (only UI is not implemented) (#288).
 - New data is available in DB: estimated reading time, language, preview picture URL, article creation date.

Issues:
 - Nav bar is not well themed (selected item color, title with logo color/background).

Issue #346, opened 2016-11-06
add option to sync only on WiFi


Strubbl:
With all the automatic sync options (and the image sync coming in future), i would love to see an option in the app to synchronize only on WiFi connection and not on mobile data plan.

1. case: maybe a general option to only ever sync on wifi
2. case: maybe an option to sync on wifi only if the automatic syncing is active
3. case: maybe an option to sync on wifi only if the image download is active

Which one do you think is best? Or should we go for all three (which would add a lot of complexity for sure)?

Strubbl:
android docu for this topic https://developer.android.com/training/monitoring-device-state/connectivity-monitoring.html

Eisfreak7:
Different usecase: I'd like to be able to sync only on a specific wifi (since the wallabag instance runs locally).
One solution would be to remove [this check](https://github.com/wallabag/android-app/blob/12e87460ba62ba646944841cf7a0b57b5b47fe10/app/src/main/java/fr/gaulupeau/apps/Poche/events/EventProcessor.java#L85), which should (I hope) allow sync to be triggered by Tasker or similar apps.
That would be very flexible and easy to implement, although maybe annoying for users that don't already have Tasker or something similar installed.

di72nn:
@Eisfreak7 does any other app implement something like that? I wonder if there is a good way to do it.

Eisfreak7:
Yes, [taskwarrior-android](https://bitbucket.org/kvorobyev/taskwarriorandroid/issues/6/sync-only-in-specific-network).

Fenisu:
What about a Sync Adapter? Wouldn't that be the optimal goal?

A "system account" for wallabag, and then let Android's Doze to manage the background sync schedules, to avoid unscheduled sync wakelocks.

Strubbl:
@Fenisu sounds very cool. Can you support here with a pull request?

di72nn:
I considered Sync Adapter when I was implementing auto sync, but for some reason I decided not to use it. I don't remember why, so I'll probably need to look into it again.

Fenisu:
It might be interesting to implement it if @di72nn does not find any reason to not do it.

di72nn:
@Fenisu can you compile a list of advantages a Sync Adapter provides?

Fenisu:
[...]

PR #404, opened 2017-01-20
[WIP] Automatic Update when opening app

luvkaggarwal:
The app is automatically updated when it starts. In case there is no internet, a message is shown informing that there is no net connection, so the previous list is being loaded.

This is in response to issue no #360

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticlesListActivity.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticlesListActivity.java
@@ -73,6 +73,7 @@ public class ArticlesListActivity extends AppCompatActivity
         setContentView(R.layout.activity_articles_list);
 
         settings = new Settings(this);
+        syncQueue();
 
         adapter = new ArticlesListPagerAdapter(getSupportFragmentManager());
 
@@ -447,6 +448,7 @@ public class ArticlesListActivity extends AppCompatActivity
     private void syncQueue() {
         if(!WallabagConnection.isNetworkAvailable()) {
             Toast.makeText(this, getString(R.string.txtNetOffline), Toast.LENGTH_SHORT).show();
+            Toast.makeText(this, getString(R.string.prevLoad), Toast.LENGTH_SHORT).show();
             return;
         }
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -92,6 +92,7 @@
     <string name="articleList_ArchivedMark">Archived mark</string>
     <string name="noPreviousArticle">There is no previous article in current list</string>
     <string name="noNextArticle">There is no next article in current list</string>
+    <string name="prevLoad">Loading existing list</string>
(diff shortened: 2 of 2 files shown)


app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticlesListActivity.java
```diff
@@ -447,6 +448,7 @@ private ArticlesListFragment getFragment(int position) {
     private void syncQueue() {
         if(!WallabagConnection.isNetworkAvailable()) {
             Toast.makeText(this, getString(R.string.txtNetOffline), Toast.LENGTH_SHORT).show();
+            Toast.makeText(this, getString(R.string.prevLoad), Toast.LENGTH_SHORT).show();
```
Strubbl: Does is it make sense to have two toasts in a row?

Strubbl:
I do not want to have automatic update of article list on opening up the app due to high mobile traffice when loading the images. I would only accept it on WIFI only.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticlesListActivity.java
luvkaggarwal: Well, it can be concatenated to form one. But if we remove the second one, it would seem as if it is a bug in the app if it always pop up during the start of the app

luvkaggarwal:
You see that can be moderated as a check if the device is connected to wifi or not.
But if there is no feature of auto - update, it just comes off as a big loophole in the design
Its like you opening an app like facebook and having to press a refresh button instead of the app doing the work for you

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticlesListActivity.java
```diff
@@ -73,6 +73,7 @@ protected void onCreate(Bundle savedInstanceState) {
         setContentView(R.layout.activity_articles_list);
 
         settings = new Settings(this);
+        syncQueue();
```
di72nn: That's not what you think it is. It launches offline queue sync.

di72nn:
First of all, thanks for your contribution.

Here's my opinion regarding that feature:
 * The feature should be optional.
 * There should be an option to enable it only for unmetered networks (not critical).
 * The start logic should be more robust. The `onCreate()` method isn't suited for it. It probably should be `onStart()` with additional checks (a check for a certain timeout since last update).
 * Also I think that if there is no network, there should be no toast messages for automatic actions.

luvkaggarwal:
Thank you for your response.
I will look into it and try to implement your suggestions in the best way possible.

di72nn:
The feature was implemented in #370 (except for "only on Wi-Fi" part). Thanks for your work anyway.

Strubbl:
@di72nn  Is there a need for a new issue for the "only on Wi-Fi" part?

di72nn:
@Strubbl #346 may be updated.

Strubbl:
oh, ok. forgot about it.

PR #408, opened 2017-01-29
[WIP] Async list update

di72nn:
Need to add synchronization.


Strubbl:
this one also replaced by #874?

PR #412, opened 2017-01-29
ImageCache: ignore images with relative/incorrect URLs

di72nn:
Fixes crash in image fetching.


di72nn:
Also added support for images with relative URLs starting with `/assets/images/`.

PR #423, opened 2017-02-09
ImageCache: fix responsiveParametersPatterns

### LINKED ISSUE #422: Part of the article is not displayed
URL => https://www.adslzone.net/2017/01/09/intel-kaby-lake-procesadores-caracteristicas-precios/

Wallabag web => full article
RSS => full article
Wallabag app => just half article content (text and images)

di72nn:
Fixes #422.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/network/ImageCacheUtils.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/network/ImageCacheUtils.java
@@ -38,9 +38,9 @@ public class ImageCacheUtils {
             = Pattern.compile("<img[\\w\\W]*?src=\"([^\"]+?)\"[\\w\\W]*?>");
 
     private static final Pattern[] responsiveParametersPatterns = {
-            Pattern.compile("srcset=\".*\""),
-            Pattern.compile("sizes=\".*\""),
-            Pattern.compile("data-zoom-src=\".*\""),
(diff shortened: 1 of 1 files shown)


app/src/main/java/fr/gaulupeau/apps/Poche/network/ImageCacheUtils.java
```diff
@@ -38,9 +38,9 @@
             = Pattern.compile("<img[\\w\\W]*?src=\"([^\"]+?)\"[\\w\\W]*?>");
 
     private static final Pattern[] responsiveParametersPatterns = {
-            Pattern.compile("srcset=\".*\""),
-            Pattern.compile("sizes=\".*\""),
-            Pattern.compile("data-zoom-src=\".*\""),
+            Pattern.compile("srcset=\"[^\"]*?\""),
```
Strubbl: Good catch. my regex was the greedy one which is wrong in this case. I used the wrong quantifier. `[^\"]*?` looks complicated on the first view until i understood, that this means all chars except double quotes. `.*?` should work for the whole pattern, too, shouldn't it? But i think your expression is more explicit telling to not catch double quotes.

di72nn: Yeah, `.*?` should work too.

ngosang:
When will these changes be in Google Play?

tcitworld:
Soon.

PR #426, opened 2017-02-15
Fix Favorite and Archive icons with dark theme. Closes #415

### LINKED ISSUE #415: Favorite and Archive icons are not visible with black theme
Favorite icon is not visible with back theme. In the articles list the icon is black and should be white because the background is black.

ngosang:
Closes #415
This works for me but I'm not an Android developer, maybe there is a better way.


di72nn:
Thanks for your efforts.

There's a better way to handle it. I added a comment in #415.

I would prefer to postpone this until #370 is ready.

ngosang:
When you fix that close this PR. It's just a reminder now 😄

Strubbl:
#370 is ready now.

PR #434, opened 2017-02-21
Update translations

di72nn:
I didn't include Hungarian translation because it is at 61.2% of completion.


di72nn:
Well, there's not much to review.

PR #438, opened 2017-02-27
Fix storage access for image caching

di72nn:
It seems I've broken storage access in #433 for Android API levels ∈ [19; 22]. 
#433 was correct according to docs, but there seems to be [a bug](https://code.google.com/p/android/issues/detail?id=81357).

As a result, the app on these devices may silently ignore "cache images" option.


tcitworld:
*sigh* Google.

PR #440, opened 2017-02-28
Fixes

di72nn:
* Fix DB wipe: remove tags and reset update timestamps.
 * Fix DB migration: correctly migrate from old versions, also reset update timestamps.
 * Remove unused "author" field from Article.
 * Fix `java.lang.IllegalArgumentException: baseUrl must end in /` exception from GPlay Console.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -45,7 +45,7 @@ android {
 }
 
 greendao {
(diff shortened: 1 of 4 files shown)


tcitworld:
> Remove unused "author" field from Article.

Indeed.

PR #443, opened 2017-03-01
Use dialog instead of activity for adding new url

Sekhmet:
It is not very user-friendly and optimal to have separate activity just for one edit text and button so instead, I created floating action button, which opens dialog for it.


di72nn:
Thanks for your work.

I guess a dialog instead of activity is a good idea, but I don't think a floating action button is really needed in the app.

There are a couple of other more convenient ways to add articles (sharing from apps, adding links from other articles), so there is no need in one-tap-access of the URL adding activity/dialog from the main screen. I have no data (apart from having almost none issues reported about it), but I'll speculate that users rarely use this way of adding articles. And if it is not used, the floating button just annoys users by obstructing view and being an unnecessary visual stimulus.

nicosomb:
> I have no data (apart from having almost none issues reported about it), but I'll speculate that users rarely use this way of adding articles.

No data for me too. Just how I do with the android app: I only use the "share to wallabag" via the share menu.

tcitworld:
FYI, Pocket uses the floating action button as a search function and hides it when scrolling. They don't even have a way to add an url directly in the app through copy/paste for example.

I agree with @di72nn here.

Sekhmet:
If you think that FAB is to obstructing what about putting menu item for it back to navigation drawer?

ngosang:
I'm opposed. I like the current implementation. 👎

di72nn:
> If you think that FAB is to obstructing what about putting menu item for it back to navigation drawer?

I am for it.

> I'm opposed. I like the current implementation.

@ngosang anything specific against the dialog?

ngosang:
> @ngosang anything specific against the dialog?

go ahead!

Sekhmet:
I added the menu entry for that dialog to navigation drawer and updated my commit with your code style in XML (no space before closing tag).

di72nn:
*Please don't consider my suggestions as nit-picking, I'm just trying to have code and history clean.*

Remove commits that negate each other (add fab / remove fab; remove nav drawer entry / put nav drawer entry back; etc.).
Remove changes from `activity_main_appbar.xml` (in the end it has only formatting changes).

app/src/main/java/fr/gaulupeau/apps/Poche/ui/MainActivity.java
```diff
@@ -798,4 +799,29 @@ private void cancelConfigurationTest() {
...
+        builder.setView(view);
+
+        builder.setPositiveButton(android.R.string.ok, new DialogInterface.OnClickListener() {
+            @Override
+            public void onClick(DialogInterface dialog, int which) {
+                TextView textView = (TextView)view.findViewById(R.id.page_url);
+                ServiceHelper.addLink(getBaseContext(), textView.getText().toString());
+            }
+        });
+        builder.setNegativeButton(android.R.string.cancel, new DialogInterface.OnClickListener() {
```
di72nn: Pass `null` as a listener, it will dismiss the dialog.

Sekhmet:
Thanks for the feedback.
Is there a better way of deleting these commits than resetting to some old commit (ie. master HEAD) and creating another commit for it?

di72nn:
I didn't inspect the individual commits here, so I don't have the full picture, but I would probably create another branch and `git cherry-pick` some commits and amend them if needed.
If you have a plan in mind, you may use `git rebase -i` to do pretty much the same. You may refer to [Git - Rewriting History](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) for some instructions.
When you're done, `git push -f` the changes to the same branch.

I'd like to hear other suggestions.

Strubbl:
Or we could the use the "squash and merge" function here in Github for merging the PR

Sekhmet:
I removed unnecessary commits.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/MainActivity.java
```diff
@@ -798,4 +799,24 @@ private void cancelConfigurationTest() {
         }
     }
 
+    private void showAddBagDialog() {
+        AlertDialog.Builder builder = new AlertDialog.Builder(this);
+        builder.setTitle(getString(R.string.nav_add));
```
di72nn: There's `setTitle()` overload that accepts resource id, so `getString()` is not necessary here.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/MainActivity.java
```diff
@@ -798,4 +799,24 @@ private void cancelConfigurationTest() {
...
+
+        final View view = getLayoutInflater().inflate(R.layout.dialog_add, null);
```
di72nn: We need to suppress the lint warning like [here](https://github.com/wallabag/android-app/blob/master/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java#L437).

di72nn:
Reviewed. Sorry that I haven't said about these details earlier.

app/src/main/AndroidManifest.xml
```diff
@@ -74,9 +74,6 @@
                 <data android:mimeType="application/vnd.wap.xhtml+xml" />
             </intent-filter>
         </activity>
-        <activity
-            android:name="fr.gaulupeau.apps.Poche.ui.AddActivity"
-            android:label="@string/activity_add_label" />
```
di72nn: `@string/activity_add_label` is not used anymore. It would be a good idea to rename it to something like `@string/dialog_add_label` and use it as the dialog title (instead of `nav_add`).

di72nn:
I'm really sorry for making you go through all the hassle.

PR #447, opened 2017-03-02
Remove fitsSystemWindows property from layout with tabs

di72nn:


nicosomb:
can you send us a screenshot please?

di72nn:
Not me: I didn't have the problem with the header on my phone (probably too old Android) and I don't want to start an emulator to do this.

Strubbl:
before:

442_before

after:
442_after

Issue #445, opened 2017-03-02
Special characters not escaped in title


foxmask:
Hi,
When you save a page which contains a < and > arround a text in the title, the body is not display ; the title stops to be displayed just before the first < car

example with https://www.reddit.com/r/vuejs/comments/5x1a3t/where_to_put_template_tag/ 

gives in the title of the article "where to put "
but in the title of the window of the application I can see "where to put <...'

If needed I can provide snapshot but I think that with the reddit link you will see what's happened :)

tcitworld:
@di72nn Why the legacy label ? Can't this be closed ? You wanted to backport to 1.12 ?

di72nn:
Yeah, I was thinking about backporting it (significance to difference ratio is high, so why not?).

PR #449, opened 2017-03-02
Escape HTML-markup in title

di72nn:
#445.


tcitworld:
Isn't `Html.escapeHtml` more appropriate here ?

di72nn:
The naming seems non-obvious, but I believe `TextUtils.htmlEncode()` is more suitable for this purpose than `Html.escapeHtml()`. https://stackoverflow.com/a/35104644

PR #451, opened 2017-03-03
Remove cancel menu item from ManageArticleTagsActivity. Fixes #448

### LINKED ISSUE #448: Remove button in manage tags view
In manage tags view, if you click the "..." button only shows a "cancel" button. This is the same as clicking the "left arrow". I think you should remove the  "..." button.

screenshot_2017-03-02-20-32-25

Sekhmet:
Remove cancel menu item from ManageArticleTagsActivity.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ManageArticleTagsActivity.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ManageArticleTagsActivity.java
@@ -235,9 +235,6 @@ public class ManageArticleTagsActivity extends BaseActionBarActivity {
             case R.id.menu_manageTags_save:
                 save();
                 return true;
-            case R.id.menu_manageTags_cancel:
-                finish();
-                return true;
         }
 
         return super.onOptionsItemSelected(item);
--- a/app/src/main/res/menu/activity_manage_tags.xml
+++ b/app/src/main/res/menu/activity_manage_tags.xml
@@ -8,9 +8,5 @@
         android:title="@string/manageTags_save"
         android:icon="@drawable/ic_done_24dp"
         app:showAsAction="ifRoom" />
-    <item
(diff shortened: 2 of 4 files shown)


di72nn:
Please avoid committing changes to unrelated code.

Strubbl:
@di72nn i would be okay with that, because i see removing unnecessary newlines as an improvement. It reminds me of the [boy scout rule](http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule)

di72nn:
Sorry, my comment doesn't reflect what I had in mind. I wanted to say: *in general* avoid committing unrelated *auto-formatted* code parts unless you're sure they really improve something.

I commented because it looked to me like auto-formatting: one line moved for no particular reason, one line break removed (which breaks visual alignment of attributes). If you intentionally made these changes, I accept it.

In general I prefer to make separate commits for functional and cosmetic (code formatting, etc.) changes. In this case the changes are minor and related to the edited part, so the previous sentence doesn't apply. But if you plan to reformat a whole file, make a separate commit.

I don't remember ever thinking it through, so I might be wrong about it.

Sekhmet:
It was autoformatted because I'm used to formatting code before committing.
I think we should agree on unified code style because it's inconsistent (ie. some XML files use space before /> and others don't).

di72nn:
Unfortunately we don't have code style defined. Considering the number of contributors it was never the priority.
If someone is willing to document it - go ahead. I will contribute to it, but I have no plans starting it from scratch anytime soon.

> It was autoformatted because I'm used to formatting code before committing.

As long as we don't have a code style we agreed on, I don't see how auto-formatting unrelated code (with some unknown code style) is a good idea. Hence my initial point.

Sekhmet:
I updated it so unrelated code remains the same.

PR #452, opened 2017-03-03
update gradle

Strubbl:


di72nn:
It does complain about support library versions mismatch, doesn't it?

di72nn:
It does.
I've tried it earlier. Turns out the `compile 'com.mikepenz:aboutlibraries:5.7.3'` dependency pulls `com.android.support:cardview-v7:24.<something>` and gradle/android-plugin considers it as an error to have different versions of support libraries. I'm not sure whether it a bug or not (is [this](https://code.google.com/p/android/issues/detail?id=231324) it?), but I would suggest to postpone the update.

Anyway, I added `compile 'com.android.support:cardview-v7:25.2.0'` along the other support libs to dependencies to use explicitly the same version. It allows to build with the new gradle without errors.

Strubbl:
Oh, local in my Android studio it did not complain...

tcitworld:
Had the same issue but it didn't stopped me from building the app.

di72nn:
Yeah, but it makes linter sad (and, consequently, CI checks).

di72nn:
Consider "GradleCompatible" hint as warning until it is fixed in the android plugin.

di72nn:
Or maybe we should add `compile 'com.android.support:cardview-v7:25.3.0'` to deps instead. That way the version is overridden and there's no errors or warnings.

di72nn:
Changed my mind: added a fake dependency to explicitly set its version. The old branch is available as `update-gradle-bak`.

Strubbl:
approved from my side (technically i cannot approve via github cause it is my PR)

PR #459, opened 2017-03-06
Replace "bag it" strings

di72nn:
I has been bothering me for quite some time. I've never liked this jargon. Is it just me?

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -36,7 +36,7 @@
         <activity
             android:name="fr.gaulupeau.apps.Poche.ui.AddUrlProxyActivity"
             android:theme="@style/ProxyTheme">
-            <intent-filter android:label="@string/addActivity_label">
+            <intent-filter android:label="@string/app_name">
                 <action android:name="android.intent.action.SEND" />
 
                 <category android:name="android.intent.category.DEFAULT" />
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -34,9 +34,9 @@
         <activity
             android:name="fr.gaulupeau.apps.Poche.ui.ManageArticleTagsActivity"/>
         <activity
-            android:name="fr.gaulupeau.apps.Poche.ui.BagItProxyActivity"
+            android:name="fr.gaulupeau.apps.Poche.ui.AddUrlProxyActivity"
             android:theme="@style/ProxyTheme">
-            <intent-filter android:label="@string/label_name">
+            <intent-filter android:label="@string/addActivity_label">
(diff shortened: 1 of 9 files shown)


tcitworld:
Suits me, even though [this](https://www.wired.com/2017/02/diehard-coders-just-saved-nasas-earth-science-data/) made me think it was a good choice afterwards. But this should be a collective decision in every app.

@wallabag/contributors @wallabag/core @wallabag/translators Okay to drop the `bag it !` sentence and replace it by `add to wallabag` (or something else) ?

matteocoder:
I'm fine with it, but it might be confusing for a new user. If we want to keep it maybe we should add a little tutorial when the app starts up for the first time.

di72nn:
BTW, does any other wallabag app or add-on use "bag it"?

Rurik19:
@di72nn wallabagger has "Wallabag it!" menu item

matteocoder:
Now that I think about it, "add to wallabag" is a lot more translation friendly than "Bag it".

jlnostr:
The translation by Pocket which is just "+Pocket" is not an option IMO. "Add to wallabag" is the best string to use here, at least in my opinion.

j0k3r:
Add to wallabag is good to me.
"Bag it!" is fun, but we don't really know who's behind. What kind of service wil be used.
Also, if people put link "add to wallabag" all over the web, more people will check for that name instead of "bag" (from "bag it")

ngosang:
I prefer just "Wallabag" instead of "Add to wallabag". 

The Android share dialog looks like this:

Strubbl:
> I prefer just "Wallabag" instead of "Add to wallabag".

I agree. As long as we have only one sharing dialog option, i prefer that short variant, too.

di72nn:
I have "Add to Firefox", "Add to Pocket", "Add to Shaarli" at the top of my share list. Keep in mind that "Wallabag" would be somewhere near the bottom of the list.

ngosang:
In my Android they are sorted by use, most used at top. Anyway I don't care so much, both names are good to me.

bourvill:
Only "wallabag" sound good

nicosomb:
It's wallabag. Period. 😬

di72nn:
I take it the majority is in favor of plain "wallabag", right?

matteocoder:
The boss has spoken his will. It shall be "wallabag"

di72nn:
Updated to "wallabag". Also marked it untranslatable.

matteocoder:
@di72nn I see that there are quite a few changes in your commit, what should I do when the Italian translation gets merged?

di72nn:
@matteocoder that depends on the order in which the PRs will get merged. You'll be notified if we'll need you to change something.

Issue #288, opened 2016-08-31
Add ability to edit article title


gerroon:
Hi

It will be beneficial if the user can edit the titles and add/remove tags in the mobile app.

thanks

di72nn:
Thanks for your suggestion. Track issues #116 or #195 in regard to tags support.

di72nn:
Done.

gerroon:
thanks

PR #468, opened 2017-03-09
Add a dialog for changing article title

di72nn:
#288.


tcitworld:
Although sexy, this dialog isn't very handy with long titles since it doesn't allows viewing all of the content. Could it be possible to inject title edition in the tags management activity and make it a regular textarea ?

di72nn:
How about multi-line textarea?

tcitworld:
Works for me. I guess the title in the article should be updated right away but because of the way we do it it's necessary to reload the whole page.

di72nn:
Yeah, we probably need to show the title outside WebView as an Android View (also reading time, tags, etc.). I thought it is better to include this feature even though the WebView is not updated yet.

tcitworld:
👍

PR #498, opened 2017-04-05
Add pagination to Image Fetching

di72nn:
... to avoid loading all the articles to memory at once.


Strubbl:
can you please rebase?

app/src/main/java/fr/gaulupeau/apps/Poche/service/SecondaryService.java
```diff
@@ -243,26 +244,52 @@ private void fetchImages(ActionRequest actionRequest) {
...
+        int offset = 0;
 
-            ImageCacheUtils.cacheImages(article.getArticleId().longValue(), article.getContent());
+        while(true) {
+            Log.d(TAG, "fetchImages() looping; offset: " + offset);
 
-            updatedArticles.add(article.getArticleId());
+            List<Article> articleList = queryBuilder.list();
+
+            if(articleList.isEmpty()) break;
```
Strubbl: it would be nice to have debug message in this case, otherwise you only see the debug message "started" but nothing like "finished"

PR #510, opened 2017-04-09
Legacy updates

di72nn:
Haven't tested it yet.


di72nn:
651ef77 makes the app consider HTTP 500 error as "Unknown error" (instead of "Incorrect configuration"), that way the error notification should show more details. Fixes #517.

tcitworld:
Works as far as I can see. Didn't test on wallabag v1 servers.

PR #523, opened 2017-05-12
update to gradle 2.3.2

Strubbl:


Strubbl:
ah i remember. i already pushed this update some time ago, which failed, too

di72nn:
Show be ok now.

PR #524, opened 2017-05-17
Add Framabag preset

di72nn:
...and support for custom login page.


di72nn:
Added two minor fixes.

*\*sigh\** The logic ought to be rewritten.

PR #534, opened 2017-05-30
Make list icons theme dependent

### LINKED ISSUE #415: Favorite and Archive icons are not visible with black theme

di72nn:
Fixes #415.
Supersedes #426.


ngosang:
It works for me!

Issue #441, opened 2017-03-01
Buttons height in read article view


ngosang:
The three buttons should have the same height in read article view.
![](http://pix.toile-libre.org/upload/original/1488362989.png)

tcitworld:
(at the bottom)

di72nn:
Fixed with #537.

PR #537, opened 2017-06-01
issue#441

snufflesrea:
I remove the margin for middle button
so the height for all 3 button button will be the same height #441


ngosang:
Fixes #441. It works for me!

Strubbl:
@snufflesrea maybe you can merge @di72nn PR into yours? Then we can merge this PR.

PR #538, opened 2017-06-01
URL sharing: add an option to put title in EXTRA_TITLE, but not in EXTRA_TEXT

### LINKED ISSUE #536: Article's title set twice
Hi there,
Since #503 I've got the title set twice when sharing an article to Diaspora Native WebApp. This app use EXTRA_TEXT and EXTRA_SUBJECT. I look how Firefox for Android handle this: https://github.com/mozilla/gecko-dev/blob/1a0d9545b9805f50a70de703a3c04fc0d22e3839/mobile/android/base/java/org/mozilla/gecko/BrowserApp.java#L3624-L3627

di72nn:
Fixes #536.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
@@ -722,6 +722,14 @@ public class Settings {
         setString(R.string.pref_key_storage_dbPath, dbPath);
     }
 
+    public boolean isAddTitleToSharedTextEnabled() {
+        return getBoolean(R.string.pref_key_misc_addTitleToSharedText_enabled, true);
+    }
+
+    public void setAddTitleToSharedTextEnabled(boolean value) {
+        setBoolean(R.string.pref_key_misc_addTitleToSharedText_enabled, value);
+    }
+
     public boolean isAppendWallabagMentionEnabled() {
         return getBoolean(R.string.pref_key_misc_appendWallabagMention_enabled, true);
     }
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticleActionsHelper.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticleActionsHelper.java
@@ -22,6 +22,7 @@ import androidx.appcompat.app.AlertDialog;
 
 import fr.gaulupeau.apps.InThePoche.R;
 import fr.gaulupeau.apps.Poche.App;
+import fr.gaulupeau.apps.Poche.data.Settings;
 import fr.gaulupeau.apps.Poche.data.dao.entities.Article;
 import fr.gaulupeau.apps.Poche.service.OperationsHelper;
 import wallabag.apiwrapper.WallabagService;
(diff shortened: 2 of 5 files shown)


tcitworld:
I looked at a few apps

| App | After | Before |
| ------|---------|----------|
| Twidere | Yes | Gets twice the title |
| Facebook Messenger | No<sup>1</sup> | Gets title |
| ProtonMail | Yes | Gets twice the title |
| K-9 Mail | Yes (sets the article title as the email-title) | Yes (in title and description)
| Signal | No | Yes |
| Tusky (mastodon client) | No | Yes |
| Discord | Yes | Gets twice the title |
| Whatsapp | No <sup>1</sup> | Yes |
| Nextcloud Notes app | No | Yes |
| Weechat | No | Yes |
| Conversations | No | Yes |
| Android SMS app | No | Yes |
| Telegram | Yes | Yes |
| Open Tasks | Yes | Yes (in title and desc)


<sup>1</sup> Gets OpenGraph/webpage title data anyway

Strubbl:
Thanks for those many tests! Unfortunately the results are imbalanced. If i understand the table correctly, "Before" this PR the URL sharing worked in more cases as expected. 

▶️ As a consequence there is no need to merge this PR. Or what do you think?

Simounet:
Maybe this could be into the app preferences. Before behavior as default but the after one if using daily an app that handle information differently (just like Diaspora Native do to me).

madalu:
I would respectfully submit that URL sharing is not working correctly if it results in the title being duplicated in a receiving app. Examing the intents of Firefox, Chrome, and Youtube sugggests that they share only the url in EXTRA_TEXT and only the title in EXTRA_SUBJECT. This seems to be a common way of sharing URLs. It usually surprises me when Wallabag shares title + url to the SMS app; this is not the behavior I want or get used to when sharing from other apps, so usually have to scroll back and remove the title. An option to conform to Firefox/Chrome standards would thus be greatly appreciated. 

To add a couple of data points related, the current behavior of Wallabag results in duplicate URLs in both Markor and Orgzly. It also prevents Markor from recognizing the shared link and formatting it as Markdown. See https://github.com/gsantner/markor/issues/1401

di72nn:
@madalu thanks for bumping :)

Default is the same: add title to the text. I think duplicates are better than lost titles, but there's an option for anyone who disagrees.

Some tests:

App | On (default) | Off
--- | --- | ---
Conversations | Ok | No title
Element | Ok | No title
Signal | Ok | No title
K-9 Mail | Title is set, but also duplicated in the body | Ok (title is set, URL is body)
SMS app | Defaults to MMS: same as K-9 Mail; SMS is ok | Defaults to MMS: same as K-9 Mail; no title in SMS
Fedilab | Duplicated | Ok
Dandelion | Duplicated | Ok
Simple Notes | Ok | No title
Nextcloud Notes | Duplicated, adds all as text | Ok: adds a markdown link
Notally | Title is set, but also duplicated in the note | Ok
Markor | Duplicated | Ok: adds a markdown link
KDE Connect | Opens a text editor with the text | Opens the URL in a browser
**Total** | **Ok: 5, duplicated: 7** | **Ok: 7, title is lost: 5**

Strubbl (approved):

Issue #232, opened 2016-04-30
Preview pictures


foxmask:
When i get a article  i dont see the image that illustrates this post , image i xan see from my wallabag instance. 

Exemple of article that does not display image http://sixactualites.fr/high-tech/hubble-etonne-decouverte-dune-lune-gravitant-autour-de-planete-makemake/27331

Strubbl:
The preview picture is not part of the content of that article saved in the database. It is saved separetly (separate column in the DB). 

If we do not get the preview picture within the content of that article delivered through the RSS feed, there is no chance for the app to display that picture. 

This is not even possible with the API of Wallabag v2 by now. If you want to have that preview pic somehow included, can you please address it in the wallabag repository. Maybe the entry delivered by the API can get extended by the preview picture. 

(i hope you understand what i want to say. )

I close this issue for now, cause it is not possible to implement that at the moment. Feel free to reopen it, if we somehow can get the image from Wallabag API (and RSS feed maybe).

tcitworld:
> This is not even possible with the API of Wallabag v2 by now. If you want to have that preview pic somehow included, can you please address it in the wallabag repository. Maybe the entry delivered by the API can get extended by the preview picture. 

The API *should* return the preview picture `preview_picture`

Strubbl:
@tcitworld You are referring to the section "Adding your first entry". This is the result you should get when you add an entry. I did not test that. But i tested /api/entries.json and there the article items do not have the preview picture in Wallabag v2.0.3

tcitworld:
Not sure about it, but when you get an individual entry you should get it.

Strubbl:
It would also make sense to get when retrieving a couple of entries. The preview picture would be the only data that's missing ATM.

nicosomb:
We need to check because in the Windows Phone app (ping @jlnostr), preview pictures are available.

Strubbl:
Preview pictures (link to it) are already available via API. Sorry, I overlooked that.

jlnostr:
Well, the issue can be closed now, right?

anarcat:
@jlnostr no: the preview is in the API, but is not used yet by the app.

ngosang:
I really need this feature. There are many articles in which the most important info is in the preview_image.
Example => http://www.economist.com/blogs/graphicdetail/2017/01/daily-chart-19?fsrc=rss

Strubbl:
[...]

PR #539, opened 2017-06-01
Show preview picture in article view

di72nn:
#232.

Please somebody test image caching.


ngosang:
It works for me! Thank you

di72nn:
@ngosang if the images were downloaded previously, they won't be redownloaded just for previews. Did you try it with newly added articles?

I'll think about adding a hook to invalidate "images cached" flags for all the articles.

ngosang:
> Did you try it with newly added articles?

I added new articles but nothing.. 😞

di72nn:
Should be ok now (was caused by #543).

Strubbl:
Just tested with http://sixactualites.fr/high-tech/hubble-etonne-decouverte-dune-lune-gravitant-autour-de-planete-makemake/27331
Offline support for preview image Works okay now.

PR #545, opened 2017-06-06
adjust wallabag logo color according to theme

Strubbl:
This commit adjusts the color of the wallabag log (black or white)
according to the users selected theme. So in dark theme it is white and
in light theme (and thus in solarized) it is black.


Strubbl:
status quo before this PR:
screenshot_20170606-190038

with this PR:
screenshot_20170606-192952


Solarized and light should not have changed:

screenshot_20170606-192925


screenshot_20170606-192939

di72nn (approved):

PR #546, opened 2017-06-08
Improve image URLs regular expressions

### LINKED ISSUE #543: Image caching: singe quotation marks are not supported
Processed: ``.
Ignored: ``.

ngosang:
* Support single and double quotes. Fix #543
Example: `` or ``
* Support valid img tags with malformed URLs. URLs with singles or double quotes.
Examples: `` or ``
* Support uppercase tags. HTML is not case sensitive.
Example: `` or ``
* Support spaces between img tag and URL.
Example: ``

All of the above together

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/network/ImageCacheUtils.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/network/ImageCacheUtils.java
@@ -36,13 +36,18 @@ public class ImageCacheUtils {
 
     private static final String WALLABAG_RELATIVE_URL_PATH = "/assets/images/";
 
-    private static final Pattern IMG_URL_PATTERN
-            = Pattern.compile("<img[\\w\\W]*?src=\"([^\"]+?)\"[\\w\\W]*?>");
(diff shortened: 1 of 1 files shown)


ngosang:
@di72nn I tested it for most cases but cold you make more proves?

di72nn:
I've not tested it yet, it seems good.

Both old and new IMG_URL_PATTERNS will fail on something like `" src="hello.jpg">`, but I don't think it is important. It will require to use full-fledged HTML parser to handle all unusual cases.

Can you run a rough processing time benchmark of old vs new patterns?

ngosang:
> Both old and new IMG_URL_PATTERNS will fail on something like `" src="hello.jpg">`, but I don't think it is important. It will require to use full-fledged HTML parser to handle all unusual cases.

We can search for `src="hello.jpg"` but I think this could cause more problems than my current implementation.

> Can you run a rough processing time benchmark of old vs new patterns?

Sorry, I don't have enough skills to do that. I'm using this code in my smatphone and I didn't notice lag or more time downloading articles.

PR #585, opened 2017-06-28
solarized theme: fix icon color in action bar of article list and article view

### LINKED ISSUE #337: Some icons in Sepia theme are white with too less contrast from background
Some icons in Sepia theme are white with too less contrast from background. Either they need to be in a darker color or they should have a border like the sharing button. For details please see attached images.

the list misses a dark add and reload button:
article-list

the article view misses a dark archive and fav button and maybe a complete dark share button:
article-view

Strubbl:
This fixes #337. But there is a dependency for API level 21. I get a
lint error for the changed lines now:

> Using theme references in XML drawables requires API level 21 (current min is 11) less... (Strg+F1)
> This check scans through all the Android API calls in the application and warns about any calls that are not available on all versions targeted by this application (according to its minimum SDK attribute in the manifest).  If you really want to use this API and don't need to support older devices just set the minSdkVersion in your build.gradle or AndroidManifest.xml files.  If your code is deliberately accessing newer APIs, and you have ensured (e.g. with conditional execution) that this code will only ever be called on a supported platform, then you can annotate your class or method with the @TargetApi annotation specifying the local minimum SDK to apply, such as @TargetApi(11), such that this check considers 11 rather than your manifest file's minimum SDK as the required API level.  If you are deliberately setting android: attributes in style definitions, make sure you place this in a values-vNN folder in order to avoid running into runtime conflicts on certain devices where manufacturers have added custom attributes whose ids conflict with the new ones on later platforms.  Similarly, you can use tools:targetApi="11" in an XML file to indicate that the element will only be inflated in an adequate context.


Strubbl:
i do not know what is the impact of this change for Android API lower level 21. Can someone please check?

di72nn:
If I recall correctly, for API <21 the build system converts svg's to png's, so the changes are not applied at runtime, i.e. the icon is dark for all themes (at least on Android 4.4 (API 19)).

Strubbl:
CI also fails :|

Strubbl:
currently only main menu on article list is fixed. this needs to be done for the article view, too

Strubbl:
okay, finally the issue should be fixed and lint is also happy

Strubbl:
ping @di72nn @tcitworld for review.
i also attached screenshots of the fixes in #337

di72nn:
I'm going to review it this weekend.

Strubbl:
@di72nn did you already have a chance to review this?

di72nn:
Sorry for the delay.

Isn't it possible to use the same name for icons in `drawable` and `drawable-v21` (e.g. `ic_done_24dp`) and use that name in a common layout file instead of creating separate layouts that differ only in icon names?

Strubbl:
Thanks for the hint. I am going to try that.

Strubbl:
Okay, i am going to test a fix for #337 with different drawables instead of layouts, menus etc. I guess this will result in a completely new PR to not mess up git history too much.

Strubbl:
Just a note for me for later reference: https://developer.android.com/studio/write/vector-asset-studio.html

Strubbl:
PR #683 replaces this PR

PR #586, opened 2017-06-29
Add Android O support

tcitworld:
This adds the new [android maven repo](https://developer.android.com/preview/migration.html#uya) and sets the required version of Android from API 11 (Android 3.0) to API 14 (Android 4.0), because of a Google library. According to the Play Store stats, only 0.1% of users should be affected.

This shouldn't be released until Android O is out (in a few weeks I guess).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,10 @@ android:
     - platform-tools
 
     # The BuildTools version used by your project
-    - build-tools-25.0.3
+    - build-tools-26.0.0
 
     # The SDK version used to compile your project
-    - android-25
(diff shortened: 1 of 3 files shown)


tcitworld:
Wonder if the F-Droid guys will support the Google repo.

Strubbl:
When this has been merged, we can also update the dependency com.mikepenz:aboutlibraries to the latest release.

lokesh-krishna:
Sorry to bump but now that O has been out for a while, could this be merged?

di72nn:
Already in `master`.

PR #587, opened 2017-06-29
Multiple file formats for download

### LINKED ISSUE #286: Export in EPUB format
Now it is possible to download each article only in PDF format, but in the web app the EPUB format is available too.
It would be interesting to have this format in the android app. #

di72nn:
Resolves #286.

If the file format list needs localization, a separate labels array should be added. The order can be changed freely.

~~There are still some strings that mention PDF.~~ Made simple replacements with `%s`.

Update: for some reason my server (2.2.3) reports HTTP 500 for HTML file format. Probably unrelated to the app, haven't investigated yet.


Strubbl:
I do not have the option to export to HTML via wallabag web. Did you mean XML?

Strubbl:
Sorry, just saw that it is mentioned in the API documentation.

Strubbl:
I have tested with my wallabag instance against /api/entries/3832.html via curl calls and that works.

di72nn:
> I have tested with my wallabag instance against /api/entries/3832.html via curl calls and that works.

Does it work with the app?

Strubbl:
I should have tested this. No, does not work. I also get 500 response from server

Strubbl:
My curl was done against api endpoint /api/entries/{entry}.{_format} and not /api/entries/{entry}/export.{_format}

If i curl against the last one, i also get 500: Internal Server Error

di72nn:
`request.CRITICAL: Uncaught PHP Exception InvalidArgumentException: "The format "html" is not yet supported." at /usr/share/webapps/wallabag/src/Wallabag/CoreBundle/Helper/EntriesExport.php line 89 {"exception":"[object] (InvalidArgumentException(code: 0): The format \"html\" is not yet supported. at /usr/share/webapps/wallabag/src/Wallabag/CoreBundle/Helper/EntriesExport.php:89)"} []`

Right...

Strubbl:
Should we open an issue in wallabag core for that?

di72nn:
Probably. I've not checked 2.3 branch, but I don't remember such changes.

PR #596, opened 2017-07-01
Update readme icon

lokesh-krishna:
Updates the readme with the new launcher icon.  
Should be merged after the icon has gone live on Play Store I guess.  

I'll make a separate PR to update the screenshots to show off the improved UI and new features.


lokesh-krishna:
Updated this icon as well to the new flat icon.

lokesh-krishna:
Added credits for the icon. Is it okay the way I've done or should it be done differently?

PR #609, opened 2017-07-07
Improve CSS styles for quotes

ngosang:
This is happening only in some articles, but won't break anything. Look at the quote background.
Before:
screenshot_2017-07-07-20-52-13
With this PR:
screenshot_2017-07-07-20-53-38

### DIFF
--- a/app/src/main/assets/dark.css
+++ b/app/src/main/assets/dark.css
@@ -220,7 +220,7 @@ header.mbm {
   font-style: normal;
 }
 
-blockquote, pre {
+#article blockquote, #article pre {
   border: 1px solid #999;
   background: #222;
   padding: 0.25em;
@@ -231,6 +231,10 @@ blockquote, pre {
   font-family: monospace;
   white-space: pre;
   text-justify: none;
+}
+
+#article blockquote *, #article pre * {
+  background: inherit;
 }
 
 #article h2, #article h3, #article h4 {
--- a/app/src/main/assets/main.css
+++ b/app/src/main/assets/main.css
@@ -210,7 +210,7 @@ header.mbm {
   font-style: normal;
 }
 
-blockquote, pre {
+#article blockquote, #article pre {
   border: 1px solid #999;
   background: #DDD;
   padding: 0.25em;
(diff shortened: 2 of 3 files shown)


app/src/main/assets/dark.css
```diff
@@ -233,6 +233,11 @@ blockquote, pre {
   text-justify: none;
 }
 
+#article blockquote p, #article pre p {
+  background: inherit;
+  color: inherit;
```
di72nn: Is there a reason to set `color: inherit;`?

di72nn:
Wouldn't this
```CSS
#article blockquote *, #article pre * {
  background: inherit;
}
```
be better instead of this?
```CSS
#article blockquote p, #article pre p {
  background: inherit;
  color: inherit;
}
```

I don't know HTML and CSS well, so it's just a guess.

app/src/main/assets/dark.css
ngosang: In the first image: Text color with black background is different from the text color with gray background. It's hard to see but I have good eyes :D

ngosang:
> #article blockquote *, #article pre * {
  background: inherit;
}

That applies to all child elements, also the links. Links will be displayed just like the text, in white color.

app/src/main/assets/dark.css
di72nn: My color picker says it's `#ffffff` in both cases.

di72nn:
My suggestion doesn't contain `color: inherit;`, so only background color is affected - links are fine.

di72nn:
In response to your edits:
As I mentioned, my color picker says that `color: inherit;` have no effect in the example provided. I don't see any use for
```CSS
#article blockquote p, #article pre p {
  color: inherit;
}
```

ngosang:
Example of not using "color: inherit;"
screenshot_2017-07-09-13-54-47
Anyway, it's not important. PR rebased.

di72nn:
Thank you @ngosang. Merged.

The desired color for quoted text is black (in light themes), right?
If so, then I'm ok with adding
```CSS
#article blockquote p, #article pre p {
  color: inherit;
}
```
Feel free to open a new PR if you think we should add it.

PR #620, opened 2017-08-12
Use PAGE_UP and PAGE_DOWN keycodes for scrolling

di72nn:
Improvement for #168.

Uses parameters under "screen scrolling" in settings.

To emulate key presses (for testing):
`PAGE_UP`: `adb shell input keyevent 92`.
`PAGE_DOWN`: `adb shell input keyevent 93`.


di72nn:
Shouldn't break anything. No suggestions or improvements - merged as is.

PR #621, opened 2017-08-19
[WIP] Async lists loading

di72nn:
My second take on async loading in Android.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticleListFragment.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticleListFragment.java
@@ -151,22 +151,29 @@ public class ArticleListFragment extends RecyclerViewListFragment<Article> {
     }
 
     @Override
-    protected void resetContent() {
-        if(tagLabel != null) {
-            List<Tag> tags = tagDao.queryBuilder()
-                    .where(TagDao.Properties.Label.eq(tagLabel))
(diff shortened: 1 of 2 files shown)


ngosang:
@di72nn Why is this important? I tested this PR in Android 7.0 and I don't notice any difference.
@Strubbl why is the development too slow? Can I help in any way?

Strubbl:
async loading is interesting to include more features to the list view of the articles, e.g. adding reading time and tags to every entry.

regarding slow development: For me it is a hobby. My daily job is not Android development. I am satisfied with the current state of the app.

di72nn:
@ngosang
It is important because currently the app does a lot of work (e. g. DB querying) on the main thread. The difference should be noticeable on slow devices, and you can probably profile UI performance with some of the dev tools to see that the app is in trouble.
Adding new features (like tags) to the lists make the situation worth.

I'm not satisfied with the current state of the app and would like to improve it, but almost any issue requires diving into Android development peculiarities. Unfortunately, I'm not very interested in Android development and my free time is scarce now.

ngosang:
@Strubbl What about #671 ? It's necessary to update the @di72nn API wrapper.

Strubbl:
is this one relpaced by #874 ?

di72nn:
Unfortunately not. #874 makes loading lighter while still doing it in the UI thread, and this one is an attempt to move loading to a background thread.

Issue #497, opened 2017-04-05
Better login page detection ("Wallabag service not found" if the page minified)


di72nn:
Currently the app relies on basic string comparison to detect which page it receives. If the HTML code is modified, the app fails to recognize the page. [Here's an example of a minified login page](https://github.com/wallabag/android-app/issues/495#issuecomment-291929905).

This also applies to the API clients page parsing.

Issue #615, opened 2017-07-19
Can't connect my Wallabag instance


romu70:
Hello,
I'm self-hosting my Wallabag instance, my server runs the official Docker image, unmodified. I can use the iOS application and Wallabagger without any issue.

But, I can use this Android application. The connection assistant always returns Wallabag is not found at the specified address and I should check the URL...which is correct of course.

My Wallabag uses HTTPS with a Let's Encrypt certificate. I can create a test account of you need to diagnose.

tcitworld:
We'll need the Android and app versions.

romu70:
Device: Sony Xperia X Compact
Android 7.1.1
Wallabag 2.0.1 (105)

I'll look at this issue, but the error is not about SSL in my case.

romu70:
My server SSL diagnosis.

![screen shot 2017-07-19 at 10 02 46](https://user-images.githubusercontent.com/1898319/28356990-318c1aae-6c6a-11e7-80fb-81a1d69155ca.png)

di72nn:
May be caused by #497. Do you use some kind of page optimization/minifying?
If it didn't clear things up, can you post your server URL (here or privately on gitter)?

romu70:
Here is my login page source code:

`<!DOCTYPE html>




<html>
    <head>
                    <meta name="viewport" content="initial-scale=1.0">
            <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
            

            <link rel="apple-touch-icon" type="image/png" href="/bundles/wallabagcore/themes/_global/img/appicon/apple-touch-icon-152.png" sizes="152x152">
            <link rel="icon" type="image/png" href="/bundles/wallabagcore/themes/_global/img/appicon/apple-touch-icon-152.png" sizes="152x152">

            <link rel="apple-touch-icon" type="image/png" href="/bundles/wallabagcore/themes/_global/img/appicon/apple-touch-icon-144.png" sizes="144x144">
            <link rel="icon" type="image/png" href="/bundles/wallabagcore/themes/_global/img/appicon/apple-touch-icon-144.png" sizes="144x144">

            <link rel="apple-touch-icon" type="image/png" href="/bundles/wallabagcore/themes/_global/img/appicon/apple-touch-icon-120.png" sizes="120x120">
            <link rel="icon" type="image/png" href="/bundles/wallabagcore/themes/_global/img/appicon/apple-touch-icon-120.png" sizes="120x120">

[...]

PR #627, opened 2017-09-02
Fix login page detection when html code is minified

leblanc-simon:
Some web servers allow to minify code before send it to the client (like mod_pagespeed or or cady with minify plugin)
This merge request replace the login and csrf detection to allow use minify module in the webserver.

This merge request solve the issue #497

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagWebService.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/network/WallabagWebService.java
@@ -26,12 +26,12 @@ public class WallabagWebService {
 
     private static final String TAG = WallabagWebService.class.getSimpleName();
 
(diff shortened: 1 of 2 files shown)


di72nn:
Do you have a `mod_pagespeed` sample? It adds [additional attributes](https://github.com/wallabag/android-app/issues/615#issuecomment-322672746). I think those are not covered here.

leblanc-simon:
Yes it adds some attributes or converts some paths, but in my instance, with mod_pagespeed, this modifications works and all is OK : https://read.leblanc.io

di72nn:
@leblanc-simon thanks for your contribution (and also for clarifying the situation with `mod_pagespeed`).

I've run it against my server (no optimizations), seemed to be ok.

Not sure that we can say that this PR resolves #497 completely (it would require actual HTML parsing), but hopefully it would work in most of the cases.

I'll leave it open for now in case someone else want to test and maybe improve something.

Issue #114, opened 2015-11-12
Go fullscreen


tcitworld:

Dod9er:
+1

Lokesh-Krishna:
Is this with reference to the action bar not hiding itself when scrolling down?

tcitworld:
> Hey,
>
> have you ever thought about adding a fullscreen mode to wallabag where the upper android notification bar will be hidden?
> Firefox already offers such a mode. It improves the immersion with the app by a great deal.
>
> Love your app btw.
> Keep up the awesome work :)
>
> Greetings
> Andreas

di72nn:
Would be cool to have the title autohiding when scrolling down.

silberzwiebel:
It seems that [this stackoverflow Q&A](https://stackoverflow.com/questions/28770530/how-to-hide-actionbar-toolbar-while-scrolling-down-in-webview#35967742) is directly relevant for anyone who wants to implement the auto-hide feature.

ojob:
> Is this with reference to the action bar not hiding itself when scrolling down?

I just tried the v2.1.0-beta version, [documented to contain the implementation of this issue](https://github.com/wallabag/android-app/releases/tag/2.1.0-beta), but I still have the action bar shown. 

Is this expected/known?

Strubbl:
@ojob i guess yes, because since @di72nn 's comment there has been no activity, which has been merged

PR #635, opened 2017-09-30
add toggling touch via physical button

silberzwiebel:
Tested sucessfully on my e-ink android device (inkbook prime).

For now, `KeyEvent.KEYCODE_CAMERA` is hardcoded, as (a) this is the key used on my deice (inkbook prime) for this feature and (b) I wasn't sure how you set and retrieve app-wide options.

Also, the toast messages should be translatable, I guess, but again, I didn't know how to do this.

Feel free to change accordingly!

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
@@ -337,39 +337,42 @@ public class ReadArticleActivity extends BaseActionBarActivity {
 
     @Override
     public boolean dispatchKeyEvent(KeyEvent event) {
-        switch(event.getKeyCode()) {
-            case KeyEvent.KEYCODE_PAGE_UP:
-                scroll(true, screenScrollingPercent, smoothScrolling);
-                return true;
 
-            case KeyEvent.KEYCODE_PAGE_DOWN:
-                scroll(false, screenScrollingPercent, smoothScrolling);
-                return true;
+        // only do this once per key press (otherwise the action is done twice)
+        if (event.getAction() == KeyEvent.ACTION_DOWN) {
+            switch (event.getKeyCode()) {
+                case KeyEvent.KEYCODE_PAGE_UP:
+                    scroll(true, screenScrollingPercent, smoothScrolling);
+                    return true;
 
-            case KeyEvent.KEYCODE_CAMERA:
-                if(event.getAction() == KeyEvent.ACTION_DOWN) {
-                    // only do this once per key press (otherwise it's toggled immediately off again with the following KeyEvent.ACTION_UP
+                case KeyEvent.KEYCODE_PAGE_DOWN:
+                    scroll(false, screenScrollingPercent, smoothScrolling);
+                    return true;
+
+                case KeyEvent.KEYCODE_CAMERA:
+
                     disableTouch = !disableTouch;
 
-                    if(disableTouch) {
+                    if (disableTouch) {
                         Toast.makeText(this, "Touch screen disabled", Toast.LENGTH_SHORT).show();
                     } else {
                         Toast.makeText(this, "Touch screen enabled", Toast.LENGTH_SHORT).show();
                     }
                     Log.d(TAG, "toggling touch screen, now disableTouch is " + disableTouch);
+                    return true;
+            }
+
+
+            if (volumeButtonsScrolling) {
+                switch (event.getKeyCode()) {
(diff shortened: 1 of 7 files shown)


silberzwiebel:
Oh, this actually only blocks the touch screen that is covered by the app. Touch's on the system-wide tray bar (on top of the screen) are still processed normally. Potentially it would be useful to show wallabag in fullscreen mode?

silberzwiebel:
I just fixed another issue here: since physical keys are triggering two events (key_up and key_down), the scrolling actions were done twice. now they only happen once on the key_down event.
Should I open another pull request as it is a different issue?

Strubbl:
very cool. Thanks for contributing.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
```diff
@@ -322,25 +337,42 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+                        Toast.makeText(this, "Touch screen disabled", Toast.LENGTH_SHORT).show();
+                    } else {
+                        Toast.makeText(this, "Touch screen enabled", Toast.LENGTH_SHORT).show();
+                    }
+                    Log.d(TAG, "toggling touch screen, now disableTouch is " + disableTouch);
+                    return true;
+            }
+
+
+            if (volumeButtonsScrolling) {
```
Strubbl: maybe this if block should be merged into the switch-case block before? 
If not, we should also fix here the bug with double scrolling on one button press (handle KeyEvent.ACTION_DOWN).

Strubbl:
Fullscreen mode is already requested in issue https://github.com/wallabag/android-app/issues/114

silberzwiebel:
Thanks for the review. Actually, the volume-buttons if-block was already in the large if-block that cares for only triggering the event once per keypress (only for the key-down event). I do agree, however, that it wasn't really nice in the code. Hopefully it is easier to follow now.

Regarding fullscreen: I was able to implement this feature with a app-wide setting in https://github.com/silberzwiebel/wallabag-android/tree/fullscreenArticleView as a proof-of-concept (some smaller issues are not yet resolved).
However, since I made this no-touch pull request on my master branch, my fullscreenArticleView branch also contains the no-touch changes. I guess, the easiest then would be to merge this no-touch pull request first and then I'll open another pull request that should fix #114 
Otherwise, I could also fiddle around with my git repo to clean things, but I'd prefer not to ;).

Strubbl:
Thanks for the changes. 
The changes to localize the toast messages and making this option a setting in the app are needed to merge this PR. Since i do not have an Android dev env with me, i guess it is better get a review and maybe some support by @tcitworld , @di72nn or somebody who can test the changes. That would be awesome

silberzwiebel:
Great.

Regarding the localization of the toast messages: I'd appreciate some pointers where to add the string then (and could happily already add a German translation).

Regarding the setting: I already figured out how to do this (hopefully in a correct way), but I wondered what setting there should be: 
- enable / disable the no-touch option completely? 
- changing the physical key (I think most android devices have volume keys and a camera key, so sticking with the camera key seems reasonable and I can't think of a different useful key)?

di72nn:
@silberzwiebel thanks for your contribution. I'll try to review and test it tomorrow.

di72nn:
Created silberzwiebel/wallabag-android#1. I'm not sure what is the most desirable behaviour. Currently I implemented it like this:
 * "Disable touch" is disabled by default.
 * If you enable it in settings, then you can press the configured button in the article view to disable touch input. When you close the view and open an article again, touch input is enabled.
 * You can configure the button in settings (by pressing a button in a dialog; the dialog shows key code ~for no particular reason~ to indicate that the button press was acknowledged).

silberzwiebel:
Thanks for refactoring, I'm trying it at the moment and have some smaller comments in a moment.

Apart from that I'm happy with your implemented behavior. Only this:
> When you close the view and open an article again, touch input is enabled.

Maybe it's better to restore the previous state but acknowledge it with a toast if touch is disabled? It feels more convenient to me to not have to press the no-touch button again when I start to read.

silberzwiebel:
I've just merged into my fork the refactoring and the added features from @di72nn.
 Given that #636 is related to scrolling with the PAGE_UP / PAGE_DOWN keys but not with this no-touch feature, I think this is ready to be merged, no?

di72nn:
I still think that it is a better idea to have a separate app to block touch input, but not everyone will bother installing one. So I approve that feature.

This PR was open for almost two weeks, so I'll merge it this weekend, if no one has anything to add/say.

Issue #569, opened 2017-06-16
Reduce header height on article


pVesian:
Hi,
Since I'm used to Pocket, I'm going to suggest a feature they have: when reading an article, a tap will hide everything, the menu & even the Android notification bar. I think they do a fullscreen. That way there is much more content on screen. It reappears on a tap or when scrolling up.
Related to #442, but this is for article only.
Thanks!

Strubbl:
I guess, this is a duplicate of #114

pVesian:
#114 doesn't seem specific to an article. @tcitworld, thoughts?

tcitworld:
#114 should mostly be on article view for comfort, on the rest it's not that important.

pVesian:
@Strubbl So it appears indeed to be a duplicate. Can you please close it?

PR #642, opened 2017-10-22
fullscreen article view

### LINKED ISSUE #636: scrolling with PAGE_UP/PAGE_DOWN keys only works with second keypress
related to #620 and https://github.com/silberzwiebel/wallabag-android/pull/1#issuecomment-334997482 / #623 

After opening an article, the first keypress of a PAGE_UP or PAGE_DOWN is "catched" by the left back-button on the upper actionbar and not used to scroll the content. From the second keypress everything is working fine. Also replicable with the android emulator when sending keycodes 92 / 93 via `adb shell input keyevent 93`.

This does not happen with different keycodes (e.g. CAMERA, 27) which is kind of weird.

silberzwiebel:
This pull request implements a fullscreen article view (i.e., it Android's system bar and the SupportActionBar are hided during article reading) as suggested in #114. The feature is toggable via an app-setting (in contrast to react on a tap as suggested in #569 ).

This pull request also fixes #636, by only reacting on `KeyEvent.ACTION_UP` instead of `KeyEvent.ACTION_DOWN`, given [that `AppCompatActivity` catches the key down events.](https://android.googlesource.com/platform/frameworks/support/+/refs/heads/master/v7/appcompat/src/android/support/v7/app/AppCompatActivity.java#567)
(I'm not entirely sure whether this is the true reason, but the fix works for me.)


app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
```diff
@@ -332,7 +346,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
 
     @Override
     public boolean dispatchKeyEvent(KeyEvent event) {
-        if (event.getAction() == KeyEvent.ACTION_DOWN) {
+        if (event.getAction() == KeyEvent.ACTION_UP) {
```
di72nn: Partially breaks scrolling with volume buttons: the `ACTION_DOWN` event is not consumed by the app.

silberzwiebel: my latest commit should have fixed this issue, thanks for pointing it out.

silberzwiebel:
No pressure intended but what's your opinion on this PR? I would resolve the conflict, if you think the general idea is good.

Strubbl:
In my opinion it is nice to have that option although i think i am not going to use it on a daily basis. 

If @di72nn or @tcitworld have no issues with this, i would like to see this PR being merged asap.

di72nn:
I haven't tested it, so I *presume* it hides the title bar completely. I think hiding the title bar automatically while scrolling down should be the preferred (probably using layout features), but since there is no development in that area, I don't mind this way.

The workaround for `dispatchKeyEvent()` seems acceptable too.

silberzwiebel:
> I haven't tested it, so I presume it hides the title bar completely. 

Yes, that's correct.

> I think hiding the title bar automatically while scrolling down should be the preferred (probably using layout features), but since there is no development in that area, I don't mind this way.

Yes, that's also true. But if you don't mind, I would suggest that someone else is implementing this in another pull request, as an improvement for this feature. It seems that [this stackoverflow Q&A](https://stackoverflow.com/questions/28770530/how-to-hide-actionbar-toolbar-while-scrolling-down-in-webview#35967742) is directly relevant.

I've just resolved the conflicts.

silberzwiebel:
You are right, I messed up with git. I tried to clean it here but failed to do so. Please, head over to #678 for the clean PR and sorry for the mess!

Strubbl:
Thanks a lot for providing a cleaned up branch.

PR #643, opened 2017-10-25
update gradle to 3.0.1 and wrapper to 4.1

Strubbl:


Strubbl:
hmm, travis cannot find gradle3 ... ?

tcitworld:
See https://stackoverflow.com/questions/44406647/android-o-gradle-build-fails-with-travis-ci ?

Strubbl:
Ah cool. That might be a solution. But the stackoverflow question leaves open if gradle3 will be synced to jcenter some day.

tcitworld:
I guess the order of repositories will be used then ?

Strubbl:
I meant, let's wait until jcenter provides the gradle3. Or will that never happen?
This way we do not have to introduce another maven source then.

PR #648, opened 2017-11-01
add possibility to scroll towards the bottom to mark article as read

silberzwiebel:
this is especially useful when in no-touch mode (the button "mark article as read" can't be touched without leaving the no-touch mode). The number of times one must scroll to the bottom (and the general feature itself) should possibly be configurable. I didn't do this yet, because I wanted to hear your feedback first.


silberzwiebel:
No pressure intended but what's your opinion on this PR? I would resolve the conflict, if you think the general idea is good.

Strubbl:
Is this PR intented to be used in conjunction with #642 ? Because otherwise i am using the check mark icon in the menu bar above the article to mark an article as read (if i am too lazy to scroll to the bottom)

silberzwiebel:
No, actually it is meant for the already merged #635. Reading articles with disabled touch display also disables the possibilities to click/press/touch any other button. So, this PR implements that one can keep on pressing the physical scroll down key a number of times and then the article is marked as read (one does not need to turn touch on again).

Strubbl:
Ok, i think i understood it. With this PR, do you want to extend the feature from #635 or do you want to introduce another option, which depends on activating the option from #635?

di72nn:
The feature definitely makes sense for reading with touch disabled. Since it is not triggered for touch scrolling, it's fine by me.

silberzwiebel:
> With this PR, do you want to extend the feature from #635 or do you want to introduce another option, which depends on activating the option from #635?

I honestly don't get the difference between the two possibilities. I think it only makes sense in the no-touch mode. Do you mean whether this "automatically marking as read" feature is part of the no-touch feature, so one cannot turn it off? I think that the number of times one needs to press the scroll-down button should be configurable (but only if no-touch mode is enabled). Or does anyone think it's also useful in normal mode, where the user just can click on the button?

Strubbl:
No, i do not think this option makes sense in normal mode. It should only be triggered when toggling the physical buttons setting

di72nn:
We have an option to "scroll with volume buttons" which is available in "normal mode". So I wouldn't write this feature off as "no touch"-only.

silberzwiebel:
I've just added a preference for this feature. It is also only triggered when a physical key (volume button, page up / down) is pressed but not when scrolling via the touch screen.
So, this PR is ready to be merged (but feel free to change according to your coding style if I messed something up!)?

app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
```diff
@@ -479,6 +479,14 @@ public void setDisableTouchKeyCode(int keyCode) {
         setInt(R.string.pref_key_ui_disableTouch_keyCode, keyCode);
     }
 
+    public int getScrolledOverBottom() {
+        return getInt(R.string.pref_key_ui_scrollOverBottom, 3);
```
Strubbl: what is the hardcoded three (3) about?

silberzwiebel: That's the initial default value, so without changing the setting, you would need to scroll 3 times over the bottom to mark as read. Is there a different location where such default values should be stored?

Strubbl: good question. I do  not know. 
Isn't there a relation to this line https://github.com/wallabag/android-app/pull/648/files#diff-07a6745e4ce5bc268d522049aaabf1b3R158 ?

silberzwiebel: Quite possibly, thanks, I'll check and test this (and the other PR) on the weekend.

Strubbl: Thanks a lot for your ongoing contribution. I just had a look into the source and i saw that the location for the default value is okay like that. It is just confusing with the two different default values.

Strubbl: I can help here, so that we can merge this PR before the weekend. What should be the default value? 3 or 10?

silberzwiebel: That would be great!
Default should be 3 (imagine 10 times pushing the button for an effect :P ). Otherwise, you could also try yourself and see what feels good?

Strubbl:
I am going to merge that for now. If there are any issue i haven't seen, we need work on that afterwards in a new PR

PR #665, opened 2017-12-05
fix lint Hardcoded text

Strubbl:
fixes lint message in src/main/res/layout/list_item.xml:56: Hardcoded string "5 min", should use @string resource


di72nn:
I would rather suppress it to have "5 min" in the Android Studio preview.

Strubbl:
Thanks for that great idea, i implemented that.

PR #677, opened 2018-01-06
Fix #610

### LINKED ISSUE #610: New options for link menu: Share and Copy link
The app currently allows to open in-article links either in the browser, or add them to the Wallabag instance.

It would be good if it were possible to simply copy the link URL from there, and maybe even share it to any other willing app.

NWuensche:
Add "Copy to clipboard" and "Share" option.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
@@ -774,7 +774,9 @@ public class ReadArticleActivity extends BaseActionBarActivity {
         builder.setItems(
                 new CharSequence[]{
                         getString(R.string.d_urlAction_openInBrowser),
-                        getString(R.string.d_urlAction_addToWallabag)
+                        getString(R.string.d_urlAction_addToWallabag),
+                        getString(R.string.d_urlAction_copyToClipboard),
+                        getString(R.string.menuShare),
                 }, new DialogInterface.OnClickListener() {
                     @Override
                     public void onClick(DialogInterface dialog, int which) {
@@ -784,6 +786,16 @@ public class ReadArticleActivity extends BaseActionBarActivity {
                                 break;
                             case 1:
                                 ServiceHelper.addLink(ReadArticleActivity.this, url);
+                                break;
+                            case 2:
+                                copyOriginalURL();
+                                break;
+                            case 3:
+                                Intent sendIntent = new Intent();
+                                sendIntent.setAction(Intent.ACTION_SEND);
+                                sendIntent.putExtra(Intent.EXTRA_TEXT, url);
+                                sendIntent.setType("text/plain");
+                                startActivity(sendIntent);
                                 break;
                         }
                     }
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
@@ -776,7 +776,7 @@ public class ReadArticleActivity extends BaseActionBarActivity {
                         getString(R.string.d_urlAction_openInBrowser),
                         getString(R.string.d_urlAction_addToWallabag),
                         getString(R.string.d_urlAction_copyToClipboard),
-                        getString(R.string.menuShare),
+                        getString(R.string.menuShare)
                 }, new DialogInterface.OnClickListener() {
                     @Override
                     public void onClick(DialogInterface dialog, int which) {
@@ -788,14 +788,10 @@ public class ReadArticleActivity extends BaseActionBarActivity {
                                 ServiceHelper.addLink(ReadArticleActivity.this, url);
                                 break;
                             case 2:
-                                copyOriginalURL();
+                                copyOriginalURL(url);
                                 break;
                             case 3:
-                                Intent sendIntent = new Intent();
-                                sendIntent.setAction(Intent.ACTION_SEND);
-                                sendIntent.putExtra(Intent.EXTRA_TEXT, url);
-                                sendIntent.setType("text/plain");
-                                startActivity(sendIntent);
+                                shareArticle(url);
                                 break;
                         }
                     }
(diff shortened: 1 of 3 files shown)


app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
```diff
@@ -785,6 +787,17 @@ public void onClick(DialogInterface dialog, int which) {
...
+                            case 2:
+                                ClipboardManager clipboard = (ClipboardManager) getSystemService(Context.CLIPBOARD_SERVICE);
+                                ClipData clip = ClipData.newRawUri("wallabag", Uri.parse(url));
+                                clipboard.setPrimaryClip(clip);
+                                break;
+                            case 3:
+                                Intent sendIntent = new Intent();
+                                sendIntent.setAction(Intent.ACTION_SEND);
+                                sendIntent.putExtra(Intent.EXTRA_TEXT, url);
+                                sendIntent.setType("text/plan");
```
Strubbl: i guess it should be text/plain.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
```diff
@@ -774,7 +774,9 @@ private void handleUrlClicked(final String url) {
         builder.setItems(
                 new CharSequence[]{
                         getString(R.string.d_urlAction_openInBrowser),
-                        getString(R.string.d_urlAction_addToWallabag)
+                        getString(R.string.d_urlAction_addToWallabag),
+                        getString(R.string.d_urlAction_copytoclipboard),
```
Strubbl: please camelCase here

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
```diff
@@ -785,6 +787,17 @@ public void onClick(DialogInterface dialog, int which) {
                             case 1:
                                 ServiceHelper.addLink(ReadArticleActivity.this, url);
                                 break;
+                            case 2:
```
Strubbl: can you please adapt (rename, add parameter) and re-use the function "private void copyOriginalURL()" in this case?

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
di72nn: Also there is `shareArticle()` for this.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
```diff
@@ -785,6 +787,17 @@ public void onClick(DialogInterface dialog, int which) {
...
+                                ClipboardManager clipboard = (ClipboardManager) getSystemService(Context.CLIPBOARD_SERVICE);
+                                ClipData clip = ClipData.newRawUri("wallabag", Uri.parse(url));
+                                clipboard.setPrimaryClip(clip);
+                                break;
+                            case 3:
+                                Intent sendIntent = new Intent();
+                                sendIntent.setAction(Intent.ACTION_SEND);
+                                sendIntent.putExtra(Intent.EXTRA_TEXT, url);
+                                sendIntent.setType("text/plan");
+                                startActivity(sendIntent);
```
di72nn: Please add `break;` (it will be useful if someone adds another case without checking).

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
NWuensche: So should I leave it in or not?

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
NWuensche: Thank you. I fixed this.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
NWuensche: Thank you. I changed it.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
NWuensche: Updated it.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
Strubbl: Can you please reuse `shareArticle()` as @di72nn suggested?

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
Strubbl: `copyOriginalURL()` only copies the article's URL to clipboard, not as you wanted, the URL of the link you clicked. Can you please adapt the method (rename, add parameter for URL to copy)?

NWuensche: I tried to do it right this time.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
NWuensche: I use it now.

di72nn:
You're not supposed to copy-paste methods. You should reuse existing methods (modifying it if needed).

In this case you need to do something like this:
Modify existing `shareArticle()` to accept `articleTitle` and `articleUrl` as parameters. And probably add an overloaded version (so the old invocations would still be valid):
```Java
    private void shareArticle() {
        shareArticle(articleTitle, articleUrl);
    }

    private void shareArticle(String articleTitle, String articleUrl) {
        String shareText = articleTitle + " " + articleUrl;

        if(settings.isAppendWallabagMentionEnabled()) {
            shareText += getString(R.string.share_text_extra);
        }

        Intent send = new Intent(Intent.ACTION_SEND);
        send.setType("text/plain");
        send.putExtra(Intent.EXTRA_SUBJECT, articleTitle);
        send.putExtra(Intent.EXTRA_TEXT, shareText);

        startActivity(Intent.createChooser(send, getString(R.string.share_article_title)));
    }
```

Since the article title may be unknown, it would make sense to further modify the method to something like this:
```Java
    private void shareArticle(String articleTitle, String articleUrl) {
        String shareText = articleUrl;
        if(!TextUtils.isEmpty(articleTitle)) shareText = articleTitle + " " + shareText;

        if(settings.isAppendWallabagMentionEnabled()) {
            shareText += getString(R.string.share_text_extra);
        }

        Intent send = new Intent(Intent.ACTION_SEND);
        send.setType("text/plain");
        if(!TextUtils.isEmpty(articleTitle)) send.putExtra(Intent.EXTRA_SUBJECT, articleTitle);
        send.putExtra(Intent.EXTRA_TEXT, shareText);

        startActivity(Intent.createChooser(send, getString(R.string.share_article_title)));
    }
```

And then call it as `shareArticle(null, url)` in your code.

The same goes for `copyOriginalURL` (which is also should be renamed).

NWuensche:
@di72nn Changed it. Thank you for showing me the example. I also renamed the method.

PR #678, opened 2018-01-06
fullscreen article view (clean branch, supersedes #642)

### LINKED ISSUE #636: scrolling with PAGE_UP/PAGE_DOWN keys only works with second keypress

silberzwiebel:
This is a clean branch of #642 (see there for discussion, my first entry there follows as a quote).

> This pull request implements a fullscreen article view (i.e., it Android's system bar and the SupportActionBar are hided during article reading) as suggested in #114. The feature is toggable via an app-setting (in contrast to react on a tap as suggested in #569 ).
>
>This pull request also fixes #636, by only reacting on KeyEvent.ACTION_UP instead of KeyEvent.ACTION_DOWN, given that AppCompatActivity catches the key down events.
(I'm not entirely sure whether this is the true reason, but the fix works for me.)

The volume keys seem to rely on KeyEvent.ACTION_DOWN which is the reason for the if-else statement in the beginning of `dispatchKeyEvent()`. I successfully tested scrolling with a tablet with volume buttons and with an e-ink-reader with page-up / -down buttons.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
@@ -364,6 +364,14 @@ public class Settings {
         setBoolean(R.string.pref_key_ui_article_textAlignment_justify, value);
     }
 
+    public boolean isFullscreenArticleView() {
+        return getBoolean(R.string.pref_key_ui_article_fullscreen, true);
+    }
+
+    public void setFullScreenArticleView(boolean value) {
+        setBoolean(R.string.pref_key_ui_article_fullscreen, value);
+    }
+
     public int getReadingSpeed() {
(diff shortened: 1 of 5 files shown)


Strubbl:
From code point of view this looks good to me. But i still need to test those changes.

PR #690, opened 2018-01-26
add scrolling via hardware keys in article lists

silberzwiebel:
This PR implements scrolling of the article list(s) via hardware buttons (volume keys or page_up / page_down). I implemented this because scrolling via touch is not nice on an e-ink display. The first / last article in the list should be the last / first article after scrolling. So a user knows, where in the list he / she is while also as much articles as possible are new in the list.

Tested with my e-ink device (page up/down) and a tablet (volume keys).

(I thought this would be easier but this took me some time. Android seems to like complicating stuff, anyway ...)

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticleListsFragment.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticleListsFragment.java
@@ -219,31 +219,35 @@ public class ArticleListsFragment extends Fragment implements Sortable, Searchab
 
     public void scroll(boolean up) {
 
-        LinearLayoutManager listLayout = getCurrentFragment().recyclerViewLayoutManager;
+        ArticleListFragment currentFragment = getCurrentFragment();
 
-        int numberOfVisibleItems =
-                listLayout.findLastCompletelyVisibleItemPosition() -
-                        listLayout.findFirstCompletelyVisibleItemPosition() + 1;
+        if( currentFragment != null && currentFragment.recyclerViewLayoutManager != null) {
+            LinearLayoutManager listLayout = currentFragment.recyclerViewLayoutManager;
 
-        int oldPositionOnTop = listLayout.findFirstCompletelyVisibleItemPosition();
+            int numberOfVisibleItems =
+                    listLayout.findLastCompletelyVisibleItemPosition() -
+                            listLayout.findFirstCompletelyVisibleItemPosition() + 1;
 
-        // scroll so that as many new articles are visible than possible with one overlap
-        int newPositionOnTop;
-        if (up) {
-            newPositionOnTop = oldPositionOnTop - numberOfVisibleItems + 1;
-        } else {
-            newPositionOnTop = oldPositionOnTop + numberOfVisibleItems - 1;
+            int oldPositionOnTop = listLayout.findFirstCompletelyVisibleItemPosition();
+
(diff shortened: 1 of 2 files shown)


app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticleListsFragment.java
```diff
@@ -216,4 +217,33 @@ private void updateList(int position, boolean forceContentUpdate) {
         }
     }
 
+    public void scroll(boolean up) {
+
+        LinearLayoutManager listLayout = getCurrentFragment().recyclerViewLayoutManager;
```
di72nn: A null-check for both `getCurrentFragment()` and `recyclerViewLayoutManager` wouldn't hurt here.

di72nn:
Haven't tested it. If it works, I'm ok with it.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ArticleListsFragment.java
silberzwiebel: Done in the last commit.

di72nn:
Tested it. It have broken the "back" button (at least a physical button, not sure about the virtual one). You should do this:
```Java
    @Override
    public boolean onKeyDown(int keyCode, KeyEvent event) {
        if (currentFragment instanceof ArticleListsFragment) {
            switch (keyCode) {
                case KeyEvent.KEYCODE_PAGE_UP:
                case KeyEvent.KEYCODE_PAGE_DOWN:
                    ((ArticleListsFragment) currentFragment).scroll(keyCode == KeyEvent.KEYCODE_PAGE_UP);
                    return true;

                case KeyEvent.KEYCODE_VOLUME_UP:
                case KeyEvent.KEYCODE_VOLUME_DOWN:
                    if (settings.isVolumeButtonsScrollingEnabled()) {
                        ((ArticleListsFragment) currentFragment).scroll(keyCode == KeyEvent.KEYCODE_VOLUME_UP);
                        return true;
                    }
                    break;
            }
        }

        return super.onKeyDown(keyCode, event);
    }
```
Differences: `return true` only if you handled the event, `return super.onKeyDown(keyCode, event)` otherwise.

silberzwiebel:
Thanks for the catch, I've updated the source code like you suggested. Unfortunately, I messed up again with the git branch, so there are merge-commits from the master-branch not related to this pull request in this PR. The only change is in commit 2ddbb37.
I can provide a clean branch if necessary.

Strubbl:
Yes, please provide a clean branch and i'll merge it.

silberzwiebel:
Alright, will do so in the next couple of days.
For the future: What is the easiest way to rebase (not sure if the correct term) a branch to the current master branch (i.e., to update the commits on a PR branch forked from an old version of the master branch to be on top of the updated master branch)?  Because apparently I'm using the wrong way to do so ...

di72nn:
[`git rebase <master branch> <your feature branch>`](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) does exactly that. However, it may become complicated sometimes.

I recommend to read at least first three chapters of the [Pro Git book](https://git-scm.com/book), if you would like to confidently use git in day-to-day dev life.

silberzwiebel:
Thats it, a clean branch :) `git rebase master scrollArticleList`, two times `git rebase --skip` and `git push -f` did the trick.

PR #638, opened 2017-10-20
Update launcher icon with flat long-shadow variant

lokesh-krishna:
As discussed at #317, here's the PR to update the app launcher icon. Round icon support hasn't been added and neither has support for adaptive icons. I don't know how to do either and I think we'll need Michael Cook's help for the latter with regards to a foreground and background layer unless someone here can do it.

All the `.png`s were generated using the provided `.svg`.

Also, as mentioned at #595, the icons need go into the `mipmap` directory only if we are building different apks for different display density (which I think we aren't).

Also, the discussion at #595 mentioned the welcome icons but I don't think they need changing because they represent the service while these icons represent one specific product based off that service. Also, they look better in that context. That's just my understanding of it and I will make a PR for them as well if someone feels that is required.

Someone will need to credit Michael Cook for the icon in the About screen so that the requisite credits are present when this icon goes live. This is how he said he would like to be credited:  

> If you'd like to include credits, please do something like: Michael Cook, http://cookicons.co

Would be nice if this was also mentioned in the changelog for the version in which this goes live for that's how I discovered him in the first place and I'd like to pay that forward.

PR #714, opened 2018-05-28
Adaptive icon support

lokesh-krishna:
This adds the new icon from #638 but now with support for adaptive icons on devices running Android 8.0 or higher. The new icon in the shape of a rounded square is used as the legacy icon to be shown on devices running older versions of Android.

The icons have been moved from the `drawables` directories to `mipmap` directories as recommended [here](https://android-developers.googleblog.com/2014/10/getting-your-apps-ready-for-nexus-6-and.html).

This is what the new icon looks like on legacy devices:  

ic_launcher

The adaptive icon looks different depending upon the mask used by the launcher or selected by the user:  
screenshot_2018-05-28 adapticon 4
screenshot_2018-05-28 adapticon 5
screenshot_2018-05-28 adapticon 6
screenshot_2018-05-28 adapticon 7

As discussed earlier, we'll need to add credits for Michael Cook in About. 
I'll make a PR to update the readme as well.


Nuntius0:
The build fails because it can't find `@mipmap/ic_launcher` as specified in the manifest, because it only looks in `main/res` for mipmaps. If you move the mipmap folder there, it'll works.

lokesh-krishna:
I would like to hear what the devs think about this. Should I move the mipmap folder or would it be a better solution to clean up the folder structure?

Kdecherf:
@Strubbl any thoughts?

Strubbl:
I like it very much, that you give credits to Michael in the Readme. 

I like the approach @Nuntius0 is describing. A cleanup of the folder struture w.r.t. the icons sounds good to me. 
Let's go for that. :+1:

PR #728, opened 2018-09-11
Save socket exception in log when network shuts down

NWuensche:
Don't send an notification with the exception to the user. Instead, save the error
in the log.

This is not finished now, as there should be a notification there to inform the user of the error. 

Update: Read Message below.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/service/IntentServiceBase.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/service/IntentServiceBase.java
@@ -61,6 +61,11 @@ public abstract class IntentServiceBase extends IntentService {
                         && e.getMessage().contains("Connection timed out")) {
                     result.setErrorType(ActionResult.ErrorType.TEMPORARY);
                     handled = true;
+                } else if(e instanceof java.net.SocketException
+                        && e.getMessage() != null
+                        && e.getMessage().contains("Software caused connection abort")) {
+                    result.setErrorType(ActionResult.ErrorType.TEMPORARY);
+                    handled = true;
                 }
             }
 


NWuensche:
Maybe someone can help me on how to use the `NotificationCompat.Builder` in the catch statement, as I don't have the Context there.

app/src/main/java/fr/gaulupeau/apps/Poche/service/IntentServiceBase.java
```diff
@@ -61,6 +62,11 @@ protected ActionResult processException(Exception e, String scope) {
                         && e.getMessage().contains("Connection timed out")) {
                     result.setErrorType(ActionResult.ErrorType.TEMPORARY);
                     handled = true;
+                } else if(e instanceof java.net.SocketException
+                        && e.getMessage() != null
+                        && e.getMessage().contains("Software caused connection abort")) {
```
Strubbl: i am not a 100% sure, but comparing text seems instable and not the correct way to catch exactly this case. 
Is the error message independent from any language settings?

Strubbl: are there different types of SocketException or why do you want only the one with that specific string in it?

Strubbl: I just scrolled some lines in this file up and saw we were handling other cases live this, too.

Strubbl:
Can you please squash your two commits into one?

NWuensche:
> Can you please squash your two commits into one?

Have done this now.

app/src/main/java/fr/gaulupeau/apps/Poche/service/IntentServiceBase.java
Strubbl: No, nothing to change. I was just too fast with writing down my thoughts. I should have inspected the source code more.

Strubbl:
Thank you! :+1:

PR #730, opened 2018-09-13
Black on black consistent in dark (high contrast) theme

### LINKED ISSUE #552: Black on black not consistent in Dark (high contrast) theme
As seen here: http://imgur.com/a/cksQH
http://imgur.com/a/cksQH

The top left's black is inconsistent with the pure (AMOLED friendly) black in the rest of the screen.

NWuensche:
Fix #552 

Adjust color settings in this theme to use black/grey in the right situations for the app bar and the navigation bar. Also add a new color for the text of highlighted items in the navigation bar for this theme.

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -48,8 +48,9 @@
     <style name="DarkTheme.PopupOverlay" parent="ThemeOverlay.AppCompat.Dark"/>
 
     <style name="DarkThemeContrast" parent="DarkTheme">
-        <item name="android:background">#000000</item>
+        <item name="android:windowBackground">@color/darkContrastBlack</item>
         <item name="android:textColor">#FFFFFF</item>
+        <item name="colorPrimary">#000000</item>
     </style>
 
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -48,8 +48,9 @@
     <style name="DarkTheme.PopupOverlay" parent="ThemeOverlay.AppCompat.Dark"/>
 
     <style name="DarkThemeContrast" parent="DarkTheme">
-        <item name="android:background">#000000</item>
+        <item name="android:windowBackground">@color/darkContrastBlack</item>
         <item name="android:textColor">#FFFFFF</item>
+        <item name="colorPrimary">#000000</item>
```
Strubbl: does it make sense to use @color/darkContrastBlack instead of the hex code?

NWuensche: I thought it would be easier to change the colors afterwards when they are stored in a variable. However, I can change the hex codes to the color variables to make everything uniform.

NWuensche:
So is there a problem with the commit?

PR #732, opened 2018-09-18
Add German translation for last update string

NWuensche:


NWuensche:
Do you have any idea why it says that it finds this string twice in the file? It occurs only one time, I checked it again.

Kdecherf:
Hello @NWuensche,

It seems that this label already exists in master as per https://github.com/wallabag/android-app/blob/master/app/src/main/res/values-de/strings.xml#L272

As long as your PR adds a line at offset 230, I think Travis merges your PR into master (considering that your PR does not have any merge conflict) and let the two labels unchanged in the resulting working tree.

Could you rebase your PR against master and update your commit accordingly?

NWuensche:
I forgot to change origin to the original repo. Therefore, I always pulled master from my own fork for the last few months... Thank you for helping me, @Kdecherf! I think that this PR is useless then. If so, please just close it.

PR #736, opened 2018-09-29
WIP: added autohiding statusbar with scrollView.SetOnScrollChangeListener()

bezmi:
This addresses [#114](https://github.com/wallabag/android-app/issues/114). I have tried to add the auto-hide functionality to the action bar by using scrollView.setOnScrollChangeListener(). I tried to do it with the webViewContent, but due to its nesting within the scrollview, it was really buggy. 

The only issue I can see is that when the bar is showing/hiding, there is artificating at the bottom of  the screen.

The system status bar doesn't get hidden, but I believe that would be relatively simple to implement.

On a side note, when playing around with the fling gestures to switch articles, I found that they were pretty unreliable. Changing them to apply on the scrollView seemed to make them much more reliable but this isn't something I've tested or fully understand...I am an android development newbie.


bezmi:
Checks are failing because the method I've implemented requires API level 23 and minimum is set to 14

Strubbl:
So what can we do about the API level problem? I would not like to merge this PR until travis is happy. 

API level 14 means Android 4.0. API Level 23 would mean Android 6.0. A quick check of the stats of the installations of wallabag in Gplay reveals that about 15% of all users are using an API level below 23. This means 85% use Android API level 23 and above anyways.

From my point of view i would just increase the minimum API level. But what do the other developers think about it? //CC @wallabag/android

bezmi:
I think it would be possible to implement this with backward compatibility by extending either ScrollView or WebView and having a custom scroll listener. I tried it with webview, but the result was very buggy because the webview is nested inside a scrollview. I am a complete noob with android and java in general, so I have no idea how to do this properly.

bezmi:
@Strubbl, in the limited time I had, I just added a check of VERSION.SDK_INT to the autohide code, so it will not autohide for those on SDK < 23. I have however adjusted the onSingleTap gestures to incorporate toggling of fullscreen if the center of the screen is tapped. If you are happy with this functionality, I will go ahead and squash the commits.

bezmi:
Currently fixing two bugs - 
1) Action bar, bottom bar and padding (mark as read/unread) are not taking the theme selected (such as solarized) in article view and are plain white.

2) The extra layout_marginTop I added to give 2x actionbar size spacing from the top (so title doesn't get cut off) pushes the bottom bar off the screen. Either need to find a more elegant way to get the needed spacing so title is not covered by actionbar, or add a bottom margin to the bottom bar to counteract this...

bezmi:
@Strubbl I believe it is now ready for review. I have tested this code with the theme fixes in #730 and it fixes the action bar colour. Do you think we still need #678 after this is merged?

Overview of changes:
1. Added option to autohide system and app bar on scroll (default = on)
2. Implemented hiding of system and app bar on tap in center of screen
3. Added extra padding to the header in the CSS so the system and app bar do not overlay the displayed text.

cheywood:
Hi @di72nn and all, it would be great to get something like this merged, providing a fullscreen reading experience without losing the ability to easily bring up the UI. Was there feedback on this PR and why it wasn't merged? I gave it a quick spin just now and it seems quite good. 

Or are there other ideas about what is desired for fullscreen on scroll? #114.

cheywood:
Coming in for a re-nudge (sorry!). I'm guessing there was discussion somewhere about this not getting merged but I haven't been able to find that. @di72nn @Strubbl ?

@bezmi if this is believed ready to merge should WIP be removed?

di72nn:
Sorry, I still can't find the time to properly look into it.

I was against approaches like that, because I think it should be implemented using layouts. But seeing that no progress in the layout department has been made, approaches *like* the one in this PR may be acceptable. I haven't checked the PR though.

cheywood:
Thanks for following up. That's great, that explains the logic behind it not moving along at least.

Issue #761, opened 2019-01-31
Local HTTP connection falls on Android 9


GeorgeHahn:
# Issue details

### Duplicate?
Have you searched the issues of this repository if your issue is already known? yes

### Actual behaviour
When trying to connect to a local server (192.168.x.x), the following message is shown and connection fails:
Connection test failed
Unknown error: CLEARTEXT communication to 192.168.x.x not permitted by network security policy

### Expected behaviour
Connection succeeds. This worked up until the last app update.

# Steps to reproduce the issue
1. Test connection using latest app on Android 9
2. Observe failure

# Environment details
- **wallabag app version**: 2.1.0 (109)
- **wallabag app installation source** (e.g. Gplay, F-Droid, manual): gplay
- **Android OS version**: 9
- **Android ROM** (e.g. stock, LineageOS, SlimRom,…): stock essential ROM
- **Android hardware**: Essential PH1
- **wallabag server version**: can fill this in if required
- **Do you have Two-Factor-Authentication enabled?**: no

# Logs
Omitted, but can submit if required.

# Your experience with wallabag Android app
Have you had any luck using wallabag Android app before? (Sometimes we get tired of reading bug reports all day and a lil' positive end note does wonders)
Yes! The Wallabag Android app is fantastic! Happy user for about two years. Thank you!

tcitworld:
https://stackoverflow.com/a/50834600/10204399

MattMattV:
I experience the same issue.

# Steps to reproduce the issue
1. Try to login to my self hosted instance of wallabag via https 

# Environment details
- **wallabag app version**: 2.1.0 (109)
- **wallabag app installation source**: F-Droid
- **Android OS version**: 9
- **Android ROM**: LineageOS 16
- **Android hardware**: OnePlus X
- **wallabag server version**: 2.3.2
- **Do you have Two-Factor-Authentication enabled?**: no

andre-dierker:
Same issue here.
Android 9 does not allow to communicate in cleartext.

tcitworld:
@GeorgeHahn @MattMattV @andre-dierker Out of curiosity, could you tell us the reason why you don't have https enabled on your servers (not possible technically, didn't felt the need, don't know how…) ?

MattMattV:
Personally, https is enabled but port 80 is open to do a redirection to the port 443

andre-dierker:
[...]

Issue #756, opened 2018-12-22
Custom server CLEARTEXT issue


sztanpet:
# Issue details

Setting up a custom wallabag server does not work.

### Duplicate?
no

### Actual behaviour
- when setting up the app, I specify the domain of my custom wallabag server with https
- the error message is "Connection test failed - Unknown error: CLEARTEXT communication to ${domain} not permitted by network security policy"

### Expected behaviour
- setting up a custom server should work

# Steps to reproduce the issue
1. have android 9 pie
2. try to set up a custom wallabag server
3. instant error

# Environment details
- **wallabag app version**: 108
- **wallabag app installation source** (e.g. Gplay, F-Droid, manual): gplay
- **Android OS version**: v9 patch level dec 5
- **Android ROM** (e.g. stock, LineageOS, SlimRom,…): stock
- **Android hardware**: xiaomi mi a2
- **wallabag server version**: 2.3.5
- **Do you have Two-Factor-Authentication enabled?**: no

Originally the wallabag server did not have ssl configured. I edited the domain name in parameters.yml to include https as a scheme. The certificate is self-signed, the self-signed CA is installed on the phone and visiting the web page in chrome works fine without any certificate issues.

# Logs
## wallabag server
no connection is established

## Web server
no connection is established

diegombeltran:
I'm experiencing the same behaviour

diegombeltran:
@sztanpet FYI
https://github.com/wallabag/wallabag/issues/2273#issuecomment-456159516

sztanpet:
indeed, thank you

norweeg:
@diegombeltran that isn't a very good solution.  I can't write a docker compose file that does this fix. why is the wallabag server downgrading to http when redirecting from / to /login?  Android 9 does not like that, apparently

norweeg:
@sztanpet @diegombeltran FYI the fix referenced in https://github.com/wallabag/wallabag/issues/2273#issuecomment-456159516  is deprecated by https://github.com/wallabag/wallabag/commit/80336f77fd6ef72dafe20b35b74d74d06592c06f

jpalczewski:
[...]

PR #762, opened 2019-02-03
Allow cleartext network traffic on Android 9

GeorgeHahn:
Address https://github.com/wallabag/android-app/issues/761.

<s>Please note that this is untested - I don't have an android dev setup available at the moment. I can test this out later if desired.</s>

Edit: Tested using the build linked below - this change functions correctly.

### DIFF
--- a/app/src/main/res/xml/network_security_config.xml
+++ b/app/src/main/res/xml/network_security_config.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <network-security-config>
-    <base-config>
+    <base-config cleartextTrafficPermitted="true">
         <trust-anchors>
             <certificates src="system"/>
             <certificates src="user"/>


Strubbl:
It would be nice if somebody can test this before we merge it. I have no wallabag instance running on http only.

Who can help here?

jpalczewski:
I'll test it in a few days - I have a dockerized wallabag over http via https(traefik + letsencrypt). Unfortunately got that issue. 

By the way - is there any way to steal apk from travis,

Strubbl:
It would be nice to get an apk from CI, but i am not aware that this is anyhow possible.

GeorgeHahn:
I'd be happy to test this, but I haven't had time to setup an android build system locally.

It looks like Travis will only store artifacts if it has been connected to an S3 bucket.

tcitworld:
To me it makes sense to enforce encrypted network traffic, since it's 2019 and Let's Encrypt makes it quite easy to add certs everywhere.

I would be fine with this option being set programmatically through settings, but it doesn't seem to be possible.

However I won't oppose this, especially if https://github.com/wallabag/android-app/issues/761#issuecomment-463092250 give me good reasons.

tcitworld:
You can test this PR by installing the debug APK (it should install alongside your production wallabag app) available at https://framadrop.org/r/Eu3as-Ug1r#5+Mzf+1axX+r0fAAgPZ/FCCRNg04wyr8DoEORuPge+w=

GeorgeHahn:
Confirmed - that build works for me. Thanks for building that!

SkYNewZ:
I confirm that work for me too!

GeorgeHahn:
Any blockers to getting this merged?

jpalczewski:
I linked to this PR from another open-source app, but I'll add comment here too - this may involve some in-depth analysis of both projects source code, but maybe someone is interested. 

Wallabag doesn't allow to even login into server, when firefly app allows to login and do some actions, only sometimes refusing due to network policy. That weird, because settings(both webapps as docker images behind traefik that handles LE certificates) is same in two situations. 

Maybe it is related with different languages, maybe it may be a bug in Android itself, or it allows to communicate to cleartext-over-TLS cases through different APIs.

@tcitworld maybe a simple warning when tested(not string written by user, but resulting URL after possible redirections server-side) server address is really clear-text only may be a good idea while disabling system-provided security.

Strubbl:
From my point of view we can merge this. @tcitworld had some doubts w.r.t. security i think. 

@tcitworld what do you think? i would like to merge this PR

Strubbl:
@jpalczewski a traefik solution to make wallabag app can be found here: https://github.com/wallabag/android-app/issues/756#issuecomment-462996975

norweeg:
instead of allowing cleartext network traffic, why not fix the web application and have it honor the [X-Forwarded-Proto header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) when redirecting requests?

GeorgeHahn:
This change brings Android 9 behavior in line with the behavior of Android <8. This PR is primarily focused on solving #761 - allowing the app to be used for people who host servers on their local networks.

It may be worth having a conversation about explicitly disallowing unencrypted communications, but I believe that conversation is out of scope for this PR & the accompanying issue.

Strubbl:
ping @tcitworld for my comment above https://github.com/wallabag/android-app/pull/762#issuecomment-467158856

tcitworld:
Let's merge this. I'd like a warning on the connection assistant when using cleartext though.

PR #763, opened 2019-02-12
update to okhttp 3.13.1

Strubbl:


Strubbl:
problem is min-api is not 26 yet:
```

> Task :app:mergeExtDexDebug
AGPBI: {"kind":"error","text":"Invoke-customs are only supported starting with Android O (--min-api 26)","sources":[{}],"tool":"D8"}
AGPBI: {"kind":"error","text":"Static interface methods are only supported starting with Android N (--min-api 24): okhttp3.Request okhttp3.Authenticator.lambda$static$0(okhttp3.Route, okhttp3.Response)","sources":[{}],"tool":"D8"}
AGPBI: {"kind":"error","text":"Static interface methods are only supported starting with Android N (--min-api 24): okhttp3.Request okhttp3.Authenticator.lambda$static$0(okhttp3.Route, okhttp3.Response)","sources":[{}],"tool":"D8"}
```

Strubbl:
If we still need to support older Android version less than 5.0, we can create a legacy version with old okhttp 3.12. 

As of October 26, 2018 the number of devices with Android < API Level 21 (5.0) is 11,1%. 
Source: https://developer.android.com/about/dashboards/

From our gplay dashboard, we have total users of 3234. 109 users have Android < 5.0, which is about 3,4%. 

Conclusion:
I would like to merge this PR.

tcitworld:
> creating releases for old devices.

Can you handle this @Strubbl ?

Strubbl:
oh yes. we should update/create a legacy branch

Strubbl:
i created the branch [legacy-older-android-5.0](https://github.com/wallabag/android-app/tree/legacy-older-android-5.0) for that. How shall we promote that? Is it possible to upload legacy apks somewhere?

tcitworld:
Attaching them to the Github releases is fine by me. I'll also mention it in the release notes.

PR #775, opened 2019-03-15
Remove font-weight change for High Contrast theme

tenkabuto:
The font-weight change in the High Contrast theme seems too aggressive to me. I suggest that the changes to background color and color are sufficient for providing the High Contrast.

I've built the app with this edit. [Click here for a screenshot of it being used](https://bikeshed.party/media/d1c6bb7c33f6f0ab186a7c10c9bf3e9f93ea7c7efe84a84ae3d83da5bf401e56.png).

### DIFF
--- a/app/src/main/assets/main.css
+++ b/app/src/main/assets/main.css
@@ -20,7 +20,6 @@ body {
 
 .high-contrast {
   background: #FFF;
-  /* font-weight: 600; */
 }
 
 .serif-font {
--- a/app/src/main/assets/main.css
+++ b/app/src/main/assets/main.css
@@ -20,7 +20,7 @@ body {
 
 .high-contrast {
   background: #FFF;
-  font-weight: 600;
+  /* font-weight: 600; */
 }
 
 .serif-font {


tcitworld:
Thanks! Could you remove the line instead of commenting it ?

tcitworld:
Thanks! Could you remove the line instead of commenting it ?

tenkabuto:
@tcitworld Yes, I'll do that soon. 👍

tenkabuto:
@tcitworld I made the changes! :fireworks:

Strubbl:
How is the text looking with a `font-weight: 500;`? (400 is the default normal font weight) Is there any difference to normal?
I think for a contrast theme a slight weighter font might make sense.

tenkabuto:
@Strubbl Good idea. I didn't consider that. [Here's a comparison photo for 400, 500, and 600.](https://drive.google.com/file/d/16FSxlllOt5gmj6fmwGFqixof3NgM8PN5/view?usp=sharing) I think the 500 does indeed look better for a high contrast theme.

Strubbl:
Thanks for the comparison. What do you think about the 500 variant?

I am not a user of this theme. If i was, i would prefer that variant for the better contrast compared to 400.

Strubbl:
Sorry, overread your opinion on the 500. 
So, let's go for that?

di72nn:
I believe the font weight was added for a reason: #84 - it probably makes some difference on e-ink screens. And if it does, maybe the font weight should be made optional somehow.

@shtrom any opinion on this?

shtrom:
Hum. I didn't remember even doing this. My nook is long gone, so I cannot test. I don't have much of an opinion either way. Feel free to merge!

di72nn:
Ok, thanks. I guess I'll involve people from #168 just in case.

@gerroon, @grote, @silberzwiebel, can you guys provide any insight into e-ink usage?

If nobody cares about the font weight, then it indeed makes sense to remove that to improve consistency between themes.

tcitworld:
Isn't the black theme also used on regular Android devices as an "energy saving" theme for AMOLED devices ?

di72nn:
Well, yeah. With dark themes the situation seems to be clear: the regular dark theme maybe used by everyone, the contrast dark may be especially useful for AMOLED, but the theme is as suitable for anyone else who likes pure black.

With the contrast light theme it may be a bit different: it *may be* used by people having devices with e-ink screens (but we don't usually hear about it unless there are some problems). So if the theme is used, it would be reasonable to copy the theme as "e-ink oriented" and remove the font weight from the regular contrast theme. Or something like that.

I'm just worried that there may be people whose experience may be worsened, but there's no telling whether they would be bothered to provide feedback. I probably overthink it though.

tenkabuto:
I agree that an alternate theme should be offered that includes increased fontweight, or eventually provide instead the option of increasing fontweight across themes.

I've been holding off on replying to this issue again until I've been able to try the app with the reduced fontweight. I only have the no-fontweight theme installed right now and I like it, but I'd like to test the mid-range one, too. 😅

silberzwiebel:
Thanks for pinging me! This is how I realized that changing the theme (from default to high contrast) actually gives a better reading experience on my e-ink device.

I had thoughts around "this font is too grayish" before but it never bothered me too much. I stick to high contrast on my device now, not sure what font-weight this means, I'm running wallabag app version 2.1.0 (109).

tenkabuto:
I have used the `500` font-weight, now, and it still feels too heavy to me while using San Serif typefaces.

With Serif typefaces enabled, I don't notice a significant difference --- the key difference here, between regular (non-high) contrast and high contrast when Serif is enabled, is the background color and text color, which appear to be too light with the regular contrast setting.

di72nn:
I suggest splitting current high contrast theme into two others: a high contrast theme without font-weight and another one specifically for e-ink screens. I can provide technical advice if needed.

@silberzwiebel if you feel like that should be the default theme on devices like yours, it is possible to set it as default here: https://github.com/wallabag/android-app/blob/57ec60a886b302f29b69db79259f84747eec9db8/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java#L141

silberzwiebel:
@di72nn Thanks for the pointer. However, since the manufacturer of my e-ink device seems to consider security a not-so-important issue, the android version is still stuck on 4.2. So, newer versions of wallabag won't run on that device (as android 5.0 is now the minimum required version), rendering an update of the default theme useless.

Strubbl:
@silberzwiebel you can find a legacy release "app-release-legacy.apk" attached to the release page https://github.com/wallabag/android-app/releases/tag/2.2.0

Can you install APKs directly to the e-ink device?

silberzwiebel:
Yes, I know that, thanks! I was just pointing out that it doesn't make sense to add a new default theme for this specific device to the codebase, because the next release (with that codebase) won't run on the device anyway.

tcitworld:
Closed in favour of #782

Issue #644, opened 2017-10-29
Check compatibility with API level 26+ changes


di72nn:
Since 2.0.2 the apps targets API level 26, which may have some not easily noticeable consequences (even for pre-Android-8 devices).

For example, the app may have problem [with broadcasts](https://developer.android.com/guide/components/broadcast-exceptions.html).

Here are two more links:
https://developer.android.com/about/versions/oreo/android-8.0-changes.html
https://developer.android.com/about/versions/oreo/android-8.0-migration.html

Unfortunately, I don't have time to look into it now.

Now also API 27 and 28.

Checklist:
- [x] Notification channels #794.
- [ ] Other undiscovered problems.

di72nn:
Also https://github.com/wallabag/android-app/pull/586#issuecomment-312032326

PR #794, opened 2019-06-17
Add notification channels support

di72nn:
#644
Thanks to reckless `targetSdkVersion` bumping we have notifications broken on Android 8+ (error messages in logcat, no notifications). It doesn't help with troubleshooting, I have to say.

This PR introduces notification channels support, which should remedy the situation.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/App.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/App.java
@@ -10,6 +10,7 @@ import fr.gaulupeau.apps.InThePoche.BuildConfig;
 import fr.gaulupeau.apps.Poche.data.DbConnection;
 import fr.gaulupeau.apps.Poche.data.Settings;
 import fr.gaulupeau.apps.Poche.events.EventProcessor;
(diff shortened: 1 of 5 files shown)


Strubbl:
rgd. `targetSdkVersion`, reckless but futureproof:

> November 2019: Updates to existing apps are required to target API level 28 or higher.

https://android-developers.googleblog.com/2019/02/expanding-target-api-level-requirements.html

di72nn:
Well, too bad the future-proofness cost months of brokenness.

However, I can see that bumping `targetSdkVersion` only in November 2019 *could* lead to fewer compatibility issues fixed.

Anyway, I don't mean any offense, I'm just saying that `targetSdkVersion` change should be accompanied with an adequate compatibility assessment (at least maybe some informational issues being created).

di72nn:
After some consideration I changed the initialization method (from `a one-time migration` to `run on every start`), because it better covers the cases when notification channels should be reinitialized. It still doesn't cover the case of device locale change, but I don't think it has any importance.

tcitworld:
Merged and released.

PR #804, opened 2019-08-17
Fix sort order of articles on app restart

### LINKED ISSUE #713: Wrong sort order
Hi

Using 2.0.2 app on Lollipop and 2.3.2 on the server side.

Everything seems to work except that  the sort order is shown oldest to newest when the app is open. But the newly added article (either via the app or the server) shown at the top as the first one which is expected but the second article at the top is the oldest article ever bagged.

The weird thing is that if I reverse the article sort order using the ui button, the newly added article is not show anymore however the order is almost equal to how server web ui list the articles except that the newly added article is not shown at the top.

thanks

### LINKED ISSUE #781: Sort order reverts to newest first on app restart
# Issue details

### Duplicate?
#713 also reports a problem with sort order, but I don't think it's the same issue (and frankly, I can't really follow the description in that issue), #74 also concerns sort order (was closed by the creator without explanation)

### Actual behaviour
- When articles are sorted as oldest first and the app is closed, the list reverts to newest first on the next app start. Additionally, two presses on the sort button are needed to go back (so I guess, part of the app thinks it's still in oldest first and the first button press sets it to newest, but the actual list was already in newest). App and server are up to date, the issue reproducibly occurs on two very different devices

### Expected behaviour
- The articles should stay ordered in oldest first

# Steps to reproduce the issue
1. Set the sort order to oldest first
2. Fully close the app (swiping in the recent app view or forcing close in app details)
3. Open the app again, find articles sorted in newest first, press sort button twice to go back to oldest first

# Environment details
- app 2.2.0 (Play Store)
- Device 1: OnePlus One, Android 9.0 via LineageOS
- Device 2: Samsung Galaxy Tab S2, Android 7.0 (Samsung stock)
- self-hosted server 2.3.7
- no two-factor

# Logs
not relevant

# Your experience with wallabag Android app
Pretty happy otherwise, look and feel make for a pleasant reading experience

### LINKED ISSUE #797: First Sorting only happens after activating sort button two times
# Issue details

### Duplicate?
Have you searched the issues of this repository if your issue is already known? yes

### Actual behaviour
Apps is starting and then i click on sort articles button. Nothing happens. I touch it again and the articles get sorted.

### Expected behaviour
On the first click on the button the articles should be sorted in reverse of the default.

# Steps to reproduce the issue
1. open app
2.click on sort button (and see nothing happens)
3. click on sort button again and see articles in reverse order

# Environment details
- **wallabag app version**: 2.2.2
- **wallabag app installation source** (e.g. Gplay, F-Droid, manual): fdroid
- **Android OS version**: 9
- **Android ROM** (e.g. stock, LineageOS, SlimRom,…): stock
- **Android hardware**: samsung s9
- **wallabag server version**: 2.3.8
- **Do you have Two-Factor-Authentication enabled?**: no

NWuensche:
When restarting the app, now the articles get sorted in the way they
were ordered before. This is done by not only checking the
savedInstanceState for sorting order but also the settings.

Close #797
Close #781
Close #713

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/RecyclerViewListFragment.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/RecyclerViewListFragment.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import fr.gaulupeau.apps.InThePoche.R;
+import fr.gaulupeau.apps.Poche.data.Settings;
 
 public abstract class RecyclerViewListFragment<T> extends Fragment
         implements Sortable, Searchable {
@@ -59,7 +60,16 @@ public abstract class RecyclerViewListFragment<T> extends Fragment
                 searchQuery = savedInstanceState.getString(STATE_SEARCH_QUERY);
             }
         }
-        if(sortOrder == null) sortOrder = Sortable.SortOrder.DESC;
+
+        if(sortOrder == null) {
+            // Try getting sortOrder from settings
+            if(this.getContext() != null) {
+                Settings settings = new Settings(this.getContext());
+                sortOrder = settings.getListSortOrder();
+            } else {
+                sortOrder = Sortable.SortOrder.DESC;
+            }
+        }
 
         itemList = new ArrayList<>();
 


di72nn:
Thanks for bringing attention to the issue.
The current UI implementation of the app is a mess (I mostly blame myself, but IMO Android software architecture contributes to that a lot), so the solution to this issue is not trivial. `RecyclerViewListFragment` shouldn't know that it works with articles (it also works with tags as `TagListFragment`), so it shouldn't be able to know where to get its initial state from (`settings.getListSortOrder()`). This class should only rely on parameters that are passed from its container. In #805 the `ArticleListFragment` instances get their parameters after they were created (I think previously I tried to use `PageChangeListener` in `ArticleListsFragment` for that, but evidently that approach doesn't work on initialization stages).

PR #807, opened 2019-08-19
Show nice notification when adding entry without connected server

### LINKED ISSUE #806: Adding new entry when server not configured leads to IlleagalArgumentException
# Issue details

### Duplicate?
Have you searched the issues of this repository if your issue is already known? Yes, it is not known.

### Actual behaviour
When the server is not configured and I'm adding a article via "Add new entry", I get `java.lang.IllegalArgumentException: Illegal URL: https://` as a Notification

### Expected behaviour
The notification should show that you have to configure the server instead of showing the plain Exception.

# Steps to reproduce the issue
1. Install Wallabag
2. Skip server configuration
3. Add any URL via Menu->"Add new entry"

# Environment details
- **wallabag app version**: 2.2.3
- **wallabag app installation source** (e.g. Gplay, F-Droid, manual): Gplay
- **Android OS version**: 8.1.0
- **Android ROM** (e.g. stock, LineageOS, SlimRom,…): stock
- **Android hardware**: Moto G5
- **wallabag server version**: not configured 
- **Do you have Two-Factor-Authentication enabled?**: no

# Your experience with wallabag Android app
Like it a lot, I also like the community around the app. :+1:

NWuensche:
Close #806

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/service/IntentServiceBase.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/service/IntentServiceBase.java
@@ -75,6 +75,9 @@ public abstract class IntentServiceBase extends IntentService {
             }
             // IOExceptions in most cases mean temporary error,
             // in some cases may mean that the action was completed anyway.
+        } else if(e instanceof IllegalArgumentException && !getSettings().isConfigurationOk()) {
+            result.setErrorType(ActionResult.ErrorType.INCORRECT_CONFIGURATION);
+            result.setMessage(e.toString());
         } else { // other exceptions meant to be handled outside
             result.setErrorType(ActionResult.ErrorType.UNKNOWN);
             result.setMessage(e.toString());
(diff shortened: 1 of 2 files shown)


app/src/main/java/fr/gaulupeau/apps/Poche/service/IntentServiceBase.java
```diff
@@ -75,6 +75,12 @@ protected ActionResult processException(Exception e, String scope) {
             }
             // IOExceptions in most cases mean temporary error,
             // in some cases may mean that the action was completed anyway.
+        } else if(e instanceof IllegalArgumentException) {
```
tcitworld: Maybe move the ternary directly into the `else if(…)` condition?
Something like this:
```suggestion
        } else if(e instanceof IllegalArgumentException && !getSettings().isConfigurationOk()) {
           result.setErrorType(ActionResult.ErrorType.INCORRECT_CONFIGURATION);
```

NWuensche: Thanks, I can do this. However, I don't know if the `IllegalArgumentException` occurs if and only if the server is not configured. Can you help me with this? Because then the settings check would be unnecessary.

di72nn: `IllegalArgumentException` is too broad - the settings check is necessary. And I endorse the suggested change.

NWuensche: Ok, I changed it!

Issue #619, opened 2017-08-08
Bug in text selection/text copy


ngosang:
Wallabag app master branch
Android 6.0.1
If i select some text, the copy menu looks ugly:
![screenshot_2017-08-08-22-20-40-706_fr gaulupeau apps inthepoche debug](https://user-images.githubusercontent.com/10577978/29093325-be339b56-7c89-11e7-9c93-d71118883c55.png)

NWuensche:
I will close this issue as #730 should have fixed the big black box and #808 fixed the black highlight color. If i was wrong and the issue still persists please reopen this issue or create a new issue.

PR #808, opened 2019-08-19
Make highlight color different from background color

NWuensche:
In dark mode high contrast, the background color for selected text in an
article  and the normal background were both black.
Now gray is used for selected text background.

I'm not sure if this also closes bug #619. I think the big black box was fixed 
with the other dark mode adjustments I made before and the selected text 
background is fixed with this PR.

Can also someone tell me if the used gray is to similar to the background in 
normal dark mode? And if so, which color should I use?


di72nn:
There seem to be several common approaches to highlighting, so I tried another one. I don't have personal preference.

This pull request changes | Inversion
--- | ---
Screenshot_20190820-151509_wallabag_pr | Screenshot_20190820-151802_wallabag_suggested
Screenshot_20190820-151530_wallabag_pr | Screenshot_20190820-152116_wallabag_suggested

Here's the code (I'm not very good with CSS, so...):
```CSS
::selection {
  color: #333;
  background: #FFF;
}

.high-contrast ::selection {
  color: #000;
}
```

And while we're at it, let's fix Solarized:

Current state | Inversion
--- | ---
Screenshot_20190820-151407_wallabag_current | Screenshot_20190820-151613_wallabag_suggested

Code:
```CSS
::selection {
  color: #fdf6e3; /* $base3 */
  background: #657B83; /* $base00 */
}
```

NWuensche:
Should I just copy the css into the commit and rebase or are you going to create a new PR?

di72nn:
You can just add the Solarized theme fix, but I'm not sure about the dark ones: I posted the screenshots to maybe get opinions.

tcitworld:
The "inversion" solution seems fine to me.

di72nn:
Let's go with inversion then.

NWuensche:
Everything should be ready now.

Issue #734, opened 2018-09-27
Improve readability of wide content


azrdev:
# Issue details

### Duplicate?
Have you searched the issues of this repository if your issue is already known? yes

### Actual behaviour
Look at content that is wider than the screen, e.g.:
- a wide `<pre>`(?) like <https://lkml.org/lkml/2018/9/16/167>
- tables from wikipedia like at <https://de.wikipedia.org/wiki/Worte_des_Vorsitzenden_Mao_Tsetung#Struktur_und_Inhalt>

Such content is not squeezed to fit into the screen width, instead one can scroll horizontally to see all of it.
However, the scrolling is difficult because swiping to much to the left goes to the next article.

### Expected behaviour
Either squeeze the content (which might look bad) or [offer to] disable going the the next article by swiping. Maybe there's some other solution, too.

# Environment details
- **wallabag app version**: 1.0.2 (106)
- **wallabag app installation source**: f-droid
- **Android OS version**: 7.1
- **Android ROM** (e.g. stock, LineageOS, SlimRom,…): custom, LineageOS 
- **Android hardware**: golden / Galaxy S3mini
- **wallabag server version**: 2.3.1

# Your experience with wallabag Android app
Have you had any luck using wallabag Android app before?

Yes, works great, thank you!

bezmi:
I have been wanting to implement a fix for this. I have noticed a general lack of responsiveness when dealing with the horizontal scrolling as I read a fair bit of articles with wide tables/code in them.

The lack of responsiveness in horizontal scrolling seems to come from the fact that there is a WebView nested within a ScrollView. To fix on the branch I have been playing around with, I have removed the bottom bar with the mark as read/unread and next/previous article buttons, and just have the WebView of the article as the primary scrolling view. This greatly improves the horizontal scrolling and allows moving diagonally as well.

So is the bottom bar really necessary? We have the mark as read/unread button on the top action bar already and next/previous article can be dealt with using gestures. IMO the bottom bar presents redundant functionality.

Regarding the swipe gestures, I believe double tap on the left/right for previous/next article to be a decent alternative that would probably result is less false positives.

What do the devs think? I have already implemented most of the current functionality without the nested scrollview, minus the bottom bar in my own branch. I just need to look into writing some code to enable smooth scrolling.

erdnaxeli:
[...]

PR #810, opened 2019-08-22
Add setting to disable swiping between articles

NWuensche:
Fix one of the requests in #734 .

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
@@ -388,6 +388,14 @@ public class Settings {
         setBoolean(R.string.pref_key_ui_keepScreenOn, keepScreenOn);
     }
 
+    public boolean getSwipeArticles() {
+        return getBoolean(R.string.pref_key_ui_swipeArticles, false);
+    }
+
+    public void setSwipeArticles(boolean swipeArticles) {
+        setBoolean(R.string.pref_key_ui_swipeArticles, swipeArticles);
+    }
+
     public Sortable.SortOrder getListSortOrder() {
         String sortOrderParam = getString(R.string.pref_key_ui_lists_sortOrder);
 
--- a/app/src/main/res/xml/preferences.xml
+++ b/app/src/main/res/xml/preferences.xml
@@ -166,6 +166,11 @@
                 android:summary="@string/pref_desc_ui_scrollOverBottom"
                 android:inputType="number"
                 android:defaultValue="3"/>
+            <CheckBoxPreference
+                android:key="@string/pref_key_ui_swipeArticles"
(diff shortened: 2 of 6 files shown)


di72nn:
> should be implemented like "Tap-to-scroll"

`if(!tapToScroll) return false;` at the start of the method is what I meant by that. It would be slightly better optimized and more concise (in a way).

app/src/main/java/fr/gaulupeau/apps/Poche/ui/preferences/SettingsActivity.java
```diff
@@ -103,6 +103,9 @@ protected void onCreate(Bundle savedInstanceState) {
 
         private boolean keepScreenOnChanged;
         private boolean oldkeepScreenOn;
+
+        private boolean swipeArticlesChanged;
```
di72nn: All these `*Changed` flags are meant for settings that need to do some action in `applyChanges()` which is not the case for this.

app/src/main/res/xml/preferences.xml
```diff
@@ -166,6 +166,11 @@
                 android:summary="@string/pref_desc_ui_scrollOverBottom"
                 android:inputType="number"
                 android:defaultValue="3"/>
+            <CheckBoxPreference
+                android:key="@string/pref_key_ui_swipeArticles"
+                android:title="@string/pref_name_ui_swipeArticles"
+                android:summary="@string/pref_desc_ui_swipeArticles"
+                android:defaultValue="true"/>
```
di72nn: I would rather have it disabled by default. The feature is not implemented well and may be confusing. But it is debatable.

NWuensche:
I did delete the `*Changed` Stuff and used the `tapToScroll` scheme. I will also rebase the commits if you want if you approve everything until now.

app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
```diff
@@ -388,6 +388,14 @@ public void setKeepScreenOn(boolean keepScreenOn) {
         setBoolean(R.string.pref_key_ui_keepScreenOn, keepScreenOn);
     }
 
+    public boolean getSwipeArticles() {
+        return getBoolean(R.string.pref_key_ui_swipeArticles, true);
```
di72nn: That should be `false` too

di72nn:
Sorry for nitpicking - just trying to keep the code clean and to reduce technical debt. Thank you for your continued efforts!

NWuensche:
No problem, I'm glad that you're helping me! I changed the `true` now. Is everything OK like this? If so, I would squash everything into one commit.

di72nn:
Yeah, looks good

NWuensche:
Everything should be nice now.

PR #828, opened 2019-09-02
Update README.md

Strubbl:
remove legacy chapter from readme as we do not maintain any legacy version anymore


Strubbl:
Shall we delete the two branches legacy and legacy-older-android5?

tcitworld:
I'm having second thoughts and will have a look if I can rebase latest changes on the Android 5 legacy branch.

PR #838, opened 2019-09-27
WIP: Latex-Math rendering

khdlr:
A little quality-of-life patch that I wrote for myself and decided to share.
Allows for Latex-math rendering in the article reader by linking the KaTeX-Autorender script.

Would this be useful for others as well?

Right now it points to `cdn.jsdelivr.net`, so it probably shouldn't be merged like that. What would be the proper way of bundling the rendering script into the app?


cloud-oak:
I guess ideally math blocks should be recognized by the backend and then be sent to the frontend in some standardized format...

How is this done for code blocks?

di72nn:
Thank you for sharing this!

> What would be the proper way of bundling the rendering script into the app?

I gave it a try [in this branch](https://github.com/wallabag/android-app/tree/cloud-oak-katex-rendering). I downloaded [the KaTeX release](https://github.com/KaTeX/KaTeX/releases/tag/v0.11.1), removed everything but the required files (also removed ttf and woff2 font files, since woff is supported by a wider range of Android versions). Seems to be working alright.
AFAIU, license-wise it should be enough to include their license file (which I did).

> I guess ideally math blocks should be recognized by the backend and then be sent to the frontend in some standardized format...

I guess. But currently they aren't. And MathJax support on the web [was added only yesterday](https://github.com/wallabag/wallabag/issues/1408).

Opinions?

cloud-oak:
I got the following error messages with the bundled scripts:
```
... D/WebView.onCM: Access to script at 'file:///android_asset/katex/contrib/auto-render.min.js' from origin 'null' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. @ 0: file:///android_asset/
... D/WebView.onCM: Access to script at 'file:///android_asset/katex/katex.min.js' from origin 'null' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. @ 0: file:///android_asset/
... D/WebView.onCM: Access to CSS stylesheet at 'file:///android_asset/katex/katex.min.css' from origin 'null' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. @ 0: file:///android_asset/
```

Setting `webviewsettings.setAllowUniversalAccessFromFileURLs(true);` as suggested [here](https://stackoverflow.com/questions/8648616/webview-javascript-cross-domain-from-a-local-html-file suggests). Fixed in #2fb8b83

di72nn:
Added some fixes in [the same branch](https://github.com/wallabag/android-app/tree/cloud-oak-katex-rendering).

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
```diff
@@ -513,6 +515,10 @@ private void showDisableTouchToast() {
     @SuppressLint("SetJavaScriptEnabled")
     private void initWebView() {
         webViewContent.getSettings().setJavaScriptEnabled(true);
+        if(settings.isMathRenderingEnabled()) {
+            // This is needed to avoid CORS-errors for the bundled katex-scripts
+            webViewContent.getSettings().setAllowUniversalAccessFromFileURLs(true);
```
di72nn: I'm still not sure about this.

tcitworld:
I'm not a fan of adding extra settings for delimiters, we already have too much of them. It makes no sense to go into settings each time for each article that handles katex markup differently. We should either try to detect it or handle all of the possibilities.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
di72nn: I think dropping `crossorigin="anonymous"` in `katex_part.html` should do it.

What's the version of Android you get this error with?

di72nn:
Good point.
@cloud-oak what's the use case for configurable delimiters?

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
cloud-oak: I was getting the error on Android 10. You are right though, dropping the `crossorigin="anonymous"` fixes this as well for me.

cloud-oak:
I would say the most critical delimiter that should be togglable in some way is the single `$`.

Enabling `$` as delimiters by default however destroys any article that talks about actual dollars:
image

However, many [math heavy blogs](https://www.inference.vc/deepsets-modeling-permutation-invariance/) use them out of convenience because that's the same as in plain LaTeX.

I agree that having it in the app settings is not ideal, but couldn't think of a better way to let the user choose.

di72nn:
I looked through a couple of articles that contain `$` and it seems that money-related `$` is usually preceded by a space. And the second `$` in `$ ... $` is not. So my idea is that if an article has two consecutive dollar signs preceded by spaces (as in "to $2 and then $3 and"), then it shouldn't be processed by the `$ ... $` pattern. Here's a regex for that: `\s\$[^$]+\s\$`.
It's dumb, but seems good enough for me. Even if it fails, it's not a big deal.

chexxor:
Migrated to Wallabag from Pocket last week. I read a lot of math articles and the $...$ syntax is quite difficult to read on my Android phone. I am happy to see it's not just me with the concern. Thanks for starting a PR to add this! (sadly I am unable to help push this forward now)

di72nn:
Merged as #1104. Sorry for delaying for so long.

Issue #115, opened 2015-11-12
Show notes and hightlight (annotations support)


tcitworld:
In the far future.

gokceozan:
Very important feature for me.

bdunogier:
Do we have any ideas or experiments on how this could be implemented ?

di72nn:
I can probably add annotations sync with server, but the UI part is tricky: the app relies on WebView for displaying article content, so annotations need to be added with JavaScript, which I have limited experience with.

bdunogier:
@di72nn would you point me to where the javascript black magic is taking place ? I'd like to have a look at it :)

bdunogier:
I'm browsing about that a bit. First, there doesn't seem to exist _recent_ libraries focusing on that on github. There are a few that haven't moved for 3, 4 years and seem to have issues on recent android versions.

About Javascript, I have found this exchange, directly related to our issue: https://stackoverflow.com/questions/41670806/how-to-highlight-text-permanently-on-android-webview. The main challenge seems to be the storage of the highlight's coordinates (line, character, whatever). The javascript itself looks pretty straightforward.

jalcine:
@bdunogier It looks like https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection or https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/getSelection can help here. I can try experimenting with this to see if we can get that kind of information (re: coordination location).

Gryphus1ne:
Any word on this? Real note support is the only thing keeping me from being able to really use Wallabag.

danihodovic:
This is main blocker from switching from Pocket to Wallabag.

er314:
Same here, 
not a wallabag user as the capability to make annotations is mandatory to me.

er314:
> so annotations need to be added with JavaScript, which I have limited experience with.

I think this feature on android doesn't have to follow the same visual paradigm as the wallabag webapp.
It could follow a simpler route, with :
- highlight the text selected for annotation (same rendering as text selection already in place, but with a different color)
- when a hightlighted text is clicked, show a new view (activity), which displays + allows writing/updating the annotation text

di72nn:
Anyone interested in this feature may check out https://github.com/wallabag/android-app/pull/843#issuecomment-584673766.

di72nn:
[...]

Issue #463, opened 2017-03-07
Bug in tags


ngosang:
wallabag app master branch

1) Add one tag to one article
2) Delete that article in the web interface, not in the app
3) Sweep articles in the app
4) The tag is still there in the tag list but if you click it there aren't articles

di72nn:
That's a known limitation (though, I forgot to document it anywhere).

It is fixable, but I don't want to fix it yet (it would require additional logic and DB queries - not something I want to add to *fast* update).

PR #843, opened 2019-10-02
Annotations support

di72nn:
Currently implemented:
 * Saving annotations (when updating from server).
 * Methods to work with annotations: `addAnnotation()`, `updateAnnotation()` and `deleteAnnotation()` in `OperationsHelper`.
 * Offline changes and client-to-server sync.
 * Demo that displays a JS alert with JSON representation of article annotations in the `ReadArticleActivity`. See `webview_htmlbase.html` and `JsAnnotationController` for details.

What's not done:
 * Proper testing.
 * Any UI:
   * HTML/JS-side for highlighting and interacting with annotations in text.
   * Android side for modifications (only annotation text can be modified) and maybe more if not implemented in JS.

I'm not yet sure, but it seems that annotation changes on the server side do not mark articles as modified, so **API clients never detect annotation changes**. If that's correct, it may become an impediment akin to server-side deleted articles. Except this one doesn't have a workaround.

Everyone is welcome to step in to help with UI and JS stuff.

#115.


di72nn:
BTW, if anyone is interested, here are some code snippets I ran from `ReadArticleActivicy` using debug evaluation to test stuff:
```

Strubbl:
When having annotation support, maybe we can add a menu item to the navigation drawer, which is called "Annotated" or the like. It shall filter for annotated articles. With the tabs we can then see the unread, starred or achived annotated articles.

di72nn:
It sure is possible. However, a harder part is some kind of UI for displaying and creating annotations - without it this code is useless.

di72nn:
Rebased. [Here's the old code](https://github.com/di72nn/wallabag-android-app/tree/annotations) just in case (github seems to save old commits, but I'm not bothered to check for how long).

di72nn:
Here's a **very** clunky AnnotatorJS-based UI.
Short version: yes it kinda works, you can go ahead and test it.

Since I'm not a web developer at all, it was an absolute pain to do. ~~Through suffering~~I learned a lot.

For that reason I didn't even try to implement the JS part myself, so I had to resort to some libraries.

I considered using [touch plugin](https://github.com/aron/annotator.touch.js), which only works with AnnotatorJS `1.2`. I managed to get it running, but the UI didn't work that well with our webview (which always wraps the whole page instead of scrolling - the "annotate" button always appeared at the bottom of the article). I also couldn't build a proper version of AnnotatorJS 1.2 (some AnnotatorJS build scripts didn't work properly in my environment). Also I'm not very happy with using a version that ancient.

Instead I built AnnotatorJS's [master](https://github.com/openannotation/annotator). It is not mobile friendly, so I kludged it with Android's context menu action: you select text, press "Annotate" in the context menu, *then* annotator's "annotate" icon appears. That may be improved of course.
Another problem is that annotator popups often appear partly outside of the screen, not sure what's the deal with this.

Anyway, that last solution kinda works (UI wise; hopefully non-UI parts work ok).
**So anyone interested in this feature should now have more than enough examples to properly implement the UI.**

di72nn:
Fixed "Annotate" not doing anything on real devices.

Removed that extra step: annotation editor appears directly after pressing "Annotate".
Made a lousy attempt to horizontally center annotator popups.

AnnotatorJS is built from [this branch](https://github.com/di72nn/annotator/tree/wallabag-android-app-changes) ([`npm install .`, `make`](https://github.com/openannotation/annotator/blob/master/HACKING.rst), take `pkg/annotator.min.js`).

di72nn:
Seeing the flurry of activity I presume this is the best it's gonna get (I'm certainly not improving it much more), so I'm marking annotations as an "experimental feature", disabling it by default and unlabeling PR's "work in progress" status.

di72nn:
Oh, I keep forgetting to thank @The-Compiler for #884 - remote WebView debugging was immensely useful, thank you!

Strubbl:
I have installed this branch on my phone and i am going to test the feature.

mberasategi:
THANK YOU! This is a real deal breaker. Really hope it can be made better, but **it works** is certainly enough to keep (at least me) from switching. 😍

mblennegard:
I tried this out the other day and noticed that annotations made in the web application does not show up in the Android app.

Maybe this is what is supposed to happen at the moment, but otherwise I can create a bug or feature request respectively.

Strubbl:
I am not sure, but my guess is that annotations are only synced to the app after a "full update" and not the default "fast update". It is a similar problem with the tags, e.g. #463

mblennegard:
@Strubbl You are absolutely right.
I just tried it and can confirm that a Full Update from the Android app indeed fetches the annotations made from the web application.

What's interesting is that any annotations made in the Android app syncs with the fast update, as the Android made annotations show up directly in the web application.
Or is it that any _edits_, such as setting tags, making annotations etc., made in the Android app syncs back to the DB immediately?

di72nn:
Created a new issue with some explanation: #1048.



> Or is it that any _edits_, such as setting tags, making annotations etc., made in the Android app syncs back to the DB immediately?

Changes made in the app are always synced ASAP.

PR #861, opened 2019-10-29
Update API wrapper

di72nn:
So I've finally put the API wrapper in order (javadocs and stuff).
This PR makes some adaptations for the new wrapper version and also optimizes article deletion.

I think this should be merger **after** the next release.


di72nn:
Updated wrapper to `v1.0.0-rc.3`.

di72nn:
Updated wrapper to `v2.0.0-beta.1`.
This version introduces better `NotFoundException` handling and transparent optimizations (like lighter "delete article", automatic usage of `hashedUrl` for `exists` if available).

di72nn:
Rebased and updated wrapper to `v2.0.0-beta.2`.

Issue #863, opened 2019-10-30
Wallabag crashes immediately after fresh install (Samsung Note 10+)


glego:
# Issue details

### Duplicate?
No

### Actual behaviour
The app crashes immediately after starting.

### Expected behaviour
The app should start.

### What have you tried to resolve the issue?
- Reinstall the app

# Steps to reproduce the issue
1. Install the app on a Samsung Note10+
2. Start the app

# Environment details
- **wallabag app version**: latest 2.2.4
- **wallabag app installation source** (e.g. Gplay, F-Droid, manual): Google Play Store
- **Android OS version**: 9
- **Android ROM** (e.g. stock, LineageOS, SlimRom,…): Stock
- **Android hardware**: Samsung Note 10+
- **wallabag server version**:  Version: 2.3.3
- **Do you have Two-Factor-Authentication enabled?**: no

# Logs
## adb logcat
[logcat.zip](https://github.com/wallabag/android-app/files/3790556/logcat.zip)

# Your experience with wallabag Android app
I switched my phone this week and put all the effort into getting the log file. Wallabag has become essential in organizing my reading material. I would hate to switch to another product because I'm unable to use the app! Keep up the good work!

Thanks,
Glenn

tcitworld:
Relevant logs:
```
java.lang.NoClassDefFoundError: Failed resolution of: Lnet/sqlcipher/database/SQLiteOpenHelper;
```

```
Caused by: java.lang.ClassNotFoundException: Didn't find class "net.sqlcipher.database.SQLiteOpenHelper" on path: DexPathList[[zip file "/data/app/fr.gaulupeau.apps.InThePoche-7zt_6GPSSx7cGpXYmMEW_w==/base.apk", zip file "/data/app/fr.gaulupeau.apps.InThePoche-7zt_6GPSSx7cGpXYmMEW_w==/split_config.en.apk", zip file "/data/app/fr.gaulupeau.apps.InThePoche-7zt_6GPSSx7cGpXYmMEW_w==/split_config.xxhdpi.apk"],nativeLibraryDirectories=[/data/app/fr.gaulupeau.apps.InThePoche-7zt_6GPSSx7cGpXYmMEW_w==/lib/arm64, /system/lib64]]
```

```
SQLiteLog: (28) failed to open "/storage/3A4D-7528/Android/data/fr.gaulupeau.apps.InThePoche/files/wallabag" with flag (131138) and mode_t (0) due to error (2)
[...]

Issue #856, opened 2019-10-26
syncOfflineQueue() IOException: unexpected end of stream on http://wallabag.example.de


Strubbl:
# Issue details

### Duplicate?
Have you searched the issues of this repository if your issue is already known? yes

### Actual behaviour
On opening the app the local changes do not get synced with the server. I get a generic error message notification.

### Expected behaviour
I want the local changes to be synced with the server without an error.

# Steps to reproduce the issue
1. add one article offline
1. get internet connection and let the it sync. the internet connection is so slow, that syncing offline changes is not yet finished
2. add another article offline
3. get messed up local changes

# Environment details
- **wallabag app version**: debug master branch commit 7ec239acecbeb0a2a0a8c49e3fb0aed905bd2eea
- **wallabag app installation source** (e.g. Gplay, F-Droid, manual): manual
- **Android OS version**: 9
- **Android ROM** (e.g. stock, LineageOS, SlimRom,…): stock
- **Android hardware**: Samsung S9
- **wallabag server version**: 2.3.8
- **Do you have Two-Factor-Authentication enabled?**: no

# Logs
## Logcat from the start/resume to the error:
https://gist.github.com/Strubbl/9cd4023546a21485e7e41b22a127c129

So the error starts near the end of the log: https://gist.github.com/Strubbl/9cd4023546a21485e7e41b22a127c129#file-java-io-ioexception-unexpected-end-of-stream-on-log-L506

## wallabag server
Please paste relevant wallabag server logs here (from file `<wallabag>/var/logs/prod.log`):

```
[2019-10-26 16:16:07] request.INFO: Matched route "api_post_entries". {"route":"api_post_entries","route_parameters":{"_controller":"Wallabag\\ApiBundle\\Controller\\EntryRestController::postEntriesAction","_format":"json","_route":"api_post_entries"},"r
equest_uri":"https://wallabag.example.de/api/entries.json","method":"POST"} []
[2019-10-26 16:16:08] app.ERROR: Authentication failure for user "", from IP "2a02:820d:1500:113d::::", with UA: "okhttp/4.2.2". [] []
```

Strubbl:
This is crazy. I have moved the database to the external storage and pulled it from there to my PC. Then i opened it with sqlitebrowser and the table QUEUE_ITEM is empty - not a single line. Where does this error come from?

Besides /storage/21DC-539E/Android/data/fr.gaulupeau.apps.InThePoche.debug/files/wallabag
[...]

PR #864, opened 2019-10-30
add dependency to android-database-sqlcipher

Strubbl:
this fixes the the error message:
`java.lang.NoClassDefFoundError: Failed resolution of: Lnet/sqlcipher/database/SQLiteOpenHelper;`

this error was part of #856 and #863


di72nn:
So that's not really an error, but an informational message (admittedly verbose and distracting).
The `SQLCipher` library is optional and we don't really need to encrypt the database.

Issue #854, opened 2019-10-18
The size of the text that works with a button.


saechimdaeki:
How about zoomin-out instead of resizing the letters with buttons?
 If you don't mind, I'll start in early November.

PR #873, opened 2019-11-17
Add a gesture function to the scrollview to insert the zoom-in out function. 

saechimdaeki:
Add a gesture function to the scroll view to insert the zoom-in out function.


saechimdaeki:
This pr is an attempt to make this app more convenient. #854

tcitworld:
What's `android-zoom-view.jar` ? https://github.com/Polidea/android-zoom-view ?

saechimdaeki:
@tcitworld https://code.google.com/archive/p/android-zoom-view/downloads
I did various tests using this library, but I didn't use it in the end. I forgot to delete it. I'm sorry.

saechimdaeki:
Unnecessary libraries have been removed.

di72nn:
It's not in the `build.gradle`, but the jar is still there.

saechimdaeki:
@di72nn  Sorry, I forgot to find it in Branch when I did it on another computer. It has now been deleted.

saechimdaeki:
I usually watch game community sites often.But the files on the site sometimes look too small, so I added them because I needed an extension. I'm sorry for the mess.

di72nn:
So you only need to zoom in on some particular details, not for a whole article?
Maybe it would be better to implement this as a temporary zoom that snaps back to the normal size as soon as you let it go?

saechimdaeki:
Then shall I add it as you say? If that's better and you don't mind.

di72nn:
@tcitworld @Strubbl @NWuensche do you think we need such a feature?

It seems the implementation can be pretty compact, so I have no strong feelings one way or the other.

saechimdaeki:
If you guys say it's okay, I'll try.If not, I won't. Tell me without reserve.

Strubbl:
I like the idea of the feature since i have some articles, which have graphics, where i cannot zoom in from wallabag. So i go to the original website and look for the pic.

> Maybe it would be better to implement this as a temporary zoom that snaps back to the normal size as soon as you let it go?

This sounds like a solution to the described problem.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
```diff
@@ -154,37 +153,58 @@ public void onCreate(Bundle savedInstanceState) {
 
         settings = App.getInstance().getSettings();
 
-        if(settings.isFullscreenArticleView()) {
+        if (settings.isFullscreenArticleView()) {
```
Strubbl: it would be nice to have a pull request without all the whitespace changes. Otherwise it is hard to review.
Of course, we can have whitespace changes, but please in a different PR.

app/src/main/res/values-en/strings.xml
```diff
@@ -1 +0,0 @@
-../values/strings.xml
```
Strubbl: i do not get why this file was changed. can you please explain

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
saechimdaeki: sorry .. It looks like it came from using the reformat function of the Android studio.

app/src/main/res/values-en/strings.xml
saechimdaeki: I deleted it arbitrarily because it was a string.xml file with nothing. Sorry that made it impossible to run on my computer.

Strubbl: This is a symlink. values-en/strings.xml points to ../values/strings.xml because english translation is our default language.

saechimdaeki: I put it back.

saechimdaeki: Here the TRAVIS CI shows all checks have failed. Can I get a review without this file for a while?

di72nn:
@saechimdaeki thank you for your work and your continued effort to fix it, but I think that if we decide to have this feature it would be better if I make a new PR based on yours with some changes I'd like to make. I would of course credit you for the initial idea and implementation.

@Strubbl 
> I like the idea of the feature since i have some articles, which have graphics, where i cannot zoom in from wallabag. So i go to the original website and look for the pic.

Note that this way of implementing it seems to be doing simple scaling - it scales already rendered image - so no extra pixels. Particularly for images it may be better to try to implement "tap to open" (I think there's an issue for that).

Strubbl:
> so no extra pixels

In my use case this would be okay because the image embedded to the article was scaled down. So in theory the pixels are already there if i could zoom in. I do not know if the simple scaling would only zoom into the smaller rendered image or if it zooms into the original image.

di72nn:
> would only zoom into the smaller rendered image or if it zooms into the original image

The smaller rendered one.

Strubbl:
In that case the feature is useless.

saechimdaeki:
Speaking of words, I tried.
Now let go of your finger and it will return to its original size.

PR #878, opened 2019-12-10
Decode HTML entities in title

Strubbl:
Sometimes the title contains umlauts like äöüß or other special chars,
which are encoded as HTML. This commit decodes the HTML entities of the
title before displaying it in the article list or the read article view.

E.g. `ü` is displayed instead of `&uuml;`.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/data/ListAdapter.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/data/ListAdapter.java
@@ -2,6 +2,9 @@ package fr.gaulupeau.apps.Poche.data;
 
 import android.content.Context;
 import androidx.recyclerview.widget.RecyclerView;
+
+import android.os.Build;
+import android.text.Html;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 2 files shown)


tcitworld:
Looks like a server issue. This API field should be strictly UTF8.

Strubbl:
In the wallabag web app the error is not seen. if i look at the source code in the web app, the title is UTF-8 and not HTML entities.

If you want a test article: https://www.golem.de/news/unix-artige-systeme-sicherheitsluecke-ermoeglicht-uebernahme-von-vpn-verbindung-1912-145403.html

di72nn:
The title in the article list of the web app does contain umlauts (visible via Inspection).
Even if it will be fixed on the server, I think it wouldn't hurt to have a fix in the app too.

However, the fact that the code is already present twice in the app (and also the fact that we don't need the original title with HTML escaped characters at all) suggests that `Html.fromHtml()` should be done in the `Updater`. In which case I'd prefer to have it fixed after other ready PRs are dealt with (to avoid conflicts).

Strubbl:
I am fine with that procedure. Let's leave this PR open. As a reminder, because i do not have opened a separate issue for this.

Issue #668, opened 2017-12-08
Malformed JSON


notabene:
Hi,

Since yesterday I get this error message from Android:

> com.squareup.moshi.JsonEncodingException: Use JsonReader.setLenient(true) to accept malformed JSON at path $

This happens whenever I try to do something:

* favorite
* archive
* udapte all

HTH.

notabene:
By popular demand (@Simounet ;)) here are additional information:

* I use [NoRoot Firewall](https://play.google.com/store/apps/details?id=app.greyshirts.firewall&hl=en) to secure what goes out (hello, noisy Google) but Wallabag has all lights set to green: it can circulate information with no restriction.
* The Wallabag instance I use is [wallabag.it](https://app.wallabag.it/).

Cheers.

ntnsndr:
Same thing:

> com.squareup.moshi.JsonEncodingException: Use JsonReader.setLenient(true) to accept malformed JSON at path $

My Android app hasn't been able to load articles for several days. It has worked fine in the past.

nicosomb:
I think that we have a problem on wallabag.it. Do you know when did the first issue appear?

ntnsndr:
I can't be sure, but between 2 and 5 days ago.

ntnsndr:
Another thing: On my web interface, I noticed that one article I saved (I can't remember if it was via plugin or app) got replicated over and over—about 20 times. It appears to still be getting replicated, even after I deleted earlier copies.

nicosomb:
The issue may come from wallabag.it or from the android app. 
We upgraded wallabag.it few days ago, but maybe the android app has issues (because we updated the android app few weeks ago with upcoming features).

nicosomb:
@notabene @ntnsndr could you tell me if this bug appeared when you were offline or online please?

di72nn:
I wasn't able to reproduce it, so I only have some generic thoughts:
 * Considering the replication issue, it is almost guaranteed to be a problem with `POST /api/entries`.
  The synchronization of offline changes may be triggered by other actions (like "mark as favorite"), so even if the error is shown after such actions, it still may be caused by sending new URL to the server.
 * On the other hand, the problem may be happening for any type of request, but it is just hard to notice multiple "mark as favorite" actions for the same article.
 * There is no difference between synchronization while online or after being offline. The logic is the same, it just doesn't do anything while offline.
[...]

PR #881, opened 2019-12-16
Extra error details

di72nn:
* Adds stacktrace to some error notifications.
* Adds a notification action to copy the details to clipboard.

The stacktrace information is not always useful, but it may help on a rare occasion (like in #668 it could help to at least determine which operation caused the error).


Strubbl:
very useful :+1:

PR #884, opened 2019-12-26
Make webviews inspectable for debug builds

The-Compiler:
This allows inspecting the WebViews via chrome://inspect in Chromium.


di72nn:
Thanks! I didn't know about that feature, it should be handy.

Issue #173, opened 2015-12-16
HTML link appear too bold and underlined


etiess:
Reading an article on the app is cool (by the way, great to have the "Save in Wallabag" option now!), but HTML links appear too bold and underlined. I think it could be nicer to have a different color to indicate links (not underline, and not bold).

I agree it's a matter of taste, but I have the impression that it's becoming a standard (look at Github for example).

Server: Framabag
App : 1.9.3.1

Strubbl:
It would be nice to have more concrete suggestions here. At least for the themes dark, light and solarized. 
Maybe @ngosang has an idea?

For high contrast themes in light and dark i would keep this bold and underlined, because color would decrease the contrast.

ngosang:
@etiess take a look #598

etiess:
Great, thanks @ngosang

PR #885, opened 2019-12-26
Improve link styling for high-contrast theme

The-Compiler:
Looking at #84, the high-contrast theme originally was created for eReader
devices.

On my device (an Onyx Boox Nova Pro), it works quite well, but both links (being
a rather dark blue) and normal text get rendered as black, making it impossible
to tell them apart from normal text.

This change makes links black and re-adds the underlining.

cc @shtrom

### DIFF
--- a/app/src/main/assets/main.css
+++ b/app/src/main/assets/main.css
@@ -78,6 +78,11 @@ a:hover, a:focus {
   text-decoration: none;
 }
 
+.high-contrast a, a:hover, a:focus {
+  color: #000;
+  text-decoration: underline;
+}
+
 h2:after {
   content: "";
   height: 4px;


di72nn:
We currently have a high-contrast and an "e-ink" theme, which only differ in that the "e-ink" one has [`font-weight` set](https://github.com/wallabag/android-app/blob/0534f849d21eea76ede2e530b3826f5e32feeceb/app/src/main/assets/main.css#L25).
I think it would be better to modify only the "e-ink" theme (since I have no idea what preferences other users of the high-contrast theme have), would it be ok with you?

If so, I would suggest to simply rename `weighted-font` style to something like `e-ink` ([in CSS](https://github.com/wallabag/android-app/blob/0534f849d21eea76ede2e530b3826f5e32feeceb/app/src/main/assets/main.css#L25) and [here](https://github.com/wallabag/android-app/blob/0534f849d21eea76ede2e530b3826f5e32feeceb/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java#L706)) and add your changes with that style.

The-Compiler:
Yeah, I wasn't sure about it - not sure who would use the high contrast theme without an e-ink device. If we're thinking of e.g. people with bad vision, I guess it'd make sense to have the link text in black (to have a high contrast) as well.

Some related discussions:

- #173, there @Strubbl said:

> For high contrast themes in light and dark i would keep this bold and underlined, because color would decrease the contrast.

- #775 which introduced the separate e-ink theme, since people seem to use the high contrast theme on AMOLED screens - from @tcitworld:

> Isn't the black theme also used on regular Android devices as an "energy saving" theme for AMOLED devices ?

Also pinging @tenkabuto @di72nn @silberzwiebel from there in case they have some more insights.

I personally don't care much either way - let me know how to proceed and I will :wink:

tcitworld:
I agree with suggestions from @di72nn. Let's at least make the `weighted-font` rename.

silberzwiebel:
I'm using the e-ink theme right now on an inkbook prime and can tell apart links from text just fine. Links are light-grey, text is black.

The-Compiler:
Hmm, okay, seems like this is only needed for Onyx devices then. After thinking about it some more, I'd propose this:

- Rename `weighted-font` to `e-ink` anyways (just because it makes things more clear)
- Try working around [Onyx' stupid style injection](https://github.com/wallabag/android-app/issues/168#issuecomment-569108417) (so that links are grey instead of black)
- If that's impossible, add a new "underline links" options to the settings, so this can be turned on independently from the theme and only for e-ink devices where this is needed.

How does that sound? Closing this in the meantime. I'll try to follow up with new PRs for these, but it might take a while, as I have quite a bit on my plate right now.

PR #886, opened 2019-12-26
Use eInk-theme on Onyx devices

The-Compiler:
With this PR, the eInk theme is automatically used on [Onyx Boox](https://www.boox.com/) devices (all of them are eReaders with an eInk display, running Android).


di72nn:
Thanks!

PR #903, opened 2020-01-09
Update API wrapper

di72nn:
This only fixes article sweeping, which didn't do anything (apart from logging errors) since the last API wrapper update.

Also, releasing a beta/alpha of the app really wouldn't hurt.


Strubbl:
yeah, release. @tcitworld can we pls have one?

Issue #905, opened 2020-01-19
App enters in `UnsuccesfulResponseException` notification loop


Zayon:
# Issue details

### Duplicate?
Have you searched the issues of this repository if your issue is already known? yes

### Actual behaviour

I tried to add an entry through the app, but my wallabag server crashed (error 500) because it could not insert the entry in the database due to the entry containing emojis and the database charset not being properly configured.

The app enters in a notification loop:

> Error: com.di77n.stuff.wallabag.apiwrapper.exceptions.UnsuccesfulResponseException: HTTP response: 500

Every few seconds the app retry to add the entry and re triggers the notification. This is really annoying.

The only way to make it stop is to force stop the app from the android settings.

The app becomes unusable as long as the server is not fixed.

### Expected behaviour

There should be a way to cancel an entry.

# Steps to reproduce the issue

1. Have a badly configured server (`database_charset: utf8` instead of `database_charset: utf8mb4`)
2. Add an entry that crash the server (entry with emojis [examle](https://github.com/simple-login/app/blob/master/README.md))

# Environment details
- **wallabag app version**: 2.3.0
- **wallabag app installation source** (e.g. Gplay, F-Droid, manual): F-Droid
- **Android OS version**: 9
- **Android ROM**: OxygenOS 9.0.10
- **Android hardware**: OnePlus 5
- **wallabag server version**: 2.3.8
- **Do you have Two-Factor-Authentication enabled?**: yes/no

# Logs
## wallabag server
Please paste relevant wallabag server logs here (from file `<wallabag>/var/logs/prod.log`):

Just any CRITICAL error log.

## Web server
Please paste relevant web server logs (e.g. from nginx, Apache, …) here:

```
172.18.0.9 - - [19/Jan/2020:15:56:59 +0100] "POST /api/entries.json HTTP/1.1" 500 83 "-" "okhttp/4.2.2"
```

# Your experience with wallabag Android app
Have you had any luck using wallabag Android app before? (Sometimes we get tired of reading bug reports all day and a lil' positive end note does wonders)

Can't live without it, wallabag would be unusable on mobile without it.

di72nn:
Hi, thanks for drawing attention to the problem.

To be honest I've always been reluctant to developing Android UI, I guess it's the main reason this is still not addressed.

Here are some thoughts for me or somebody else to implement:

1. Don't retry / block on failed items.
  Should be trivial for new URLs: simply add some field to `QueueItem` for filtering out the failed ones.
[...]

PR #915, opened 2020-02-06
Add basic options to dump and clean offline queue

di72nn:
This adds `Settings -> Miscellaneous -> Local changes` submenu with two actions:
 * Dump to file.
   This saves the contents of the offline queue to a file to the app directory in the external storage.
 * Remove first local queue item.
   This, as the name suggests, removes the first local queue item. It should be enough because only the first item directly blocks synchronization. If there are more than one problem items, the action can be repeated.

This is a very basic improvement for #905 (and some other issues).

These options are not meant to be an obvious fix a user: the user should report the problem (so something may be fixed (like not saveable articles)) and be instructed to use these options to make the app usable again.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/data/StorageHelper.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/data/StorageHelper.java
@@ -3,11 +3,16 @@ package fr.gaulupeau.apps.Poche.data;
 import androidx.core.content.ContextCompat;
 import android.util.Log;
 
+import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.nio.channels.FileChannel;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
 
 import fr.gaulupeau.apps.Poche.App;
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/fr/gaulupeau/apps/Poche/data/StorageHelper.java
```diff
@@ -114,4 +118,24 @@ public static boolean deleteFile(String path) {
...
+            throw new IllegalStateException("External storage is not writable!");
+        }
+
+        String path = getExternalStoragePath() + "/"
+                + "Local_changes_"
+                + new SimpleDateFormat("yyyyMMdd_HHmmss").format(new Date())
+                + ".txt";
+        File file = new File(path);
+
+        file.createNewFile();
```
Strubbl: We should check the return value here. It's false if the file already exists. This could accidentally happen if i touch two times in a second.

app/src/main/java/fr/gaulupeau/apps/Poche/data/StorageHelper.java
```diff
@@ -114,4 +118,24 @@ public static boolean deleteFile(String path) {
         return new File(path).delete();
     }
 
+    public static File dumpQueueData(String data) throws IOException {
```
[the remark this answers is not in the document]
di72nn: Not in the throws list (because it is a `RuntimeException`), but I did forget to catch it in `dumpOfflineQueue`. Fixed.

app/src/main/java/fr/gaulupeau/apps/Poche/data/StorageHelper.java
di72nn: I don't see any point in checking for the result *in this particular ad hoc solution*, since the file should exist after that call either way (and it will be simply overwritten if you call it twice in a second). I added a comment.

I considered adding milliseconds to file name, but I don't think it's important.

PR #928, opened 2020-02-28
Add Gradle Wrapper Validation Action

di72nn:
[Gradle Wrapper Validation Action](https://github.com/gradle/wrapper-validation-action).
This checks the committed `gradle-wrapper.jar` (but not the `gradlew` scripts?).
Not the same as [Verification of downloaded Gradle distributions](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:verification).

Also update gradle version in the wrapper task. Not sure if it is needed.


tcitworld:
> (but not the gradlew scripts?).

At least you can view their diff. :man_shrugging:

PR #930, opened 2020-03-06
Save article origin when bagging from other articles

### LINKED ISSUE #711: Set origin_url when adding articles from an article
When you click on a link in an article, wallabag offers you two options: _add to wallabag_ or _open in browser_.

It could be cool to forward the original article url when clicking _add to wallabag_ using the field `original_url`.

di72nn:
Fixes #711.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -53,7 +53,7 @@ android {
 }
 
 greendao {
-    schemaVersion 106
+    schemaVersion 107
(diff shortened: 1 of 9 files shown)


tcitworld:
Extra2 isn't really obvious…

di72nn:
I agree.
Generic field names are the consequence of using the same fields for different data depending on action type.
I'll see what can be done.

di72nn:
Replaced.

PR #944, opened 2020-04-01
Update fastlane changelog and add french translation

tcitworld:


fastlane/metadata/android/en-US/changelogs/220.txt
```diff
@@ -1,5 +1,6 @@
 - Basic search is replaced with full-text search. "Full update" is recommended.
 - Experimental annotations support (can be enabled in settings).
+- Launch QR code scanner for credentials setup from configuration assistant
```
di72nn, line 3: For future reference: apparently there's a 500-character (or byte?) limit on changelogs, so it will be cut now.

tcitworld: Oh thanks, Google Play has the same restriction anyway, so I'll just copy/paste in the future.

Issue #939, opened 2020-03-29
After interruption, TTS restarts at wrong place


ntnsndr:
# Issue details

### Duplicate?
Have you searched the issues of this repository if your issue is already known? yes/no

Yes, I haven't found it. I may have missed it—if so, apologies.

### Actual behaviour

When I'm listening to TTS (my main way of using the app), and there is some kind of interruption (such as a text message ping or a voice instruction from a navigation app), TTS re-starts after the interruption not where it left off but where I initially started it (usually the beginning of the article). This makes the app largely unusable, for instance, while running a navigation app and driving.

### Expected behaviour

TTS playback should continue after the interruption where it left off.

# Steps to reproduce the issue
1. Begin TTS playback
2. Receive instruction from an app like Google Maps or receive a text message
3. TTS re-starts where it first began

# Environment details
- **wallabag app version**: 2.3.0
- **wallabag app installation source** (e.g. Gplay, F-Droid, manual): F-Droid
- **Android OS version**: 9
- **Android ROM** (e.g. stock, LineageOS, SlimRom,…): LineageOS
- **Android hardware**: Moto X
- **wallabag server version**: 2.4.0-dev

# Your experience with wallabag Android app
Have you had any luck using wallabag Android app before? (Sometimes we get tired of reading bug reports all day and a lil' positive end note does wonders)

I love the app and use it daily. It is mostly an excellent experience. Thank you so much for all you do!

Strubbl:
Maybe @tyndare has an idea how to fix this?

di72nn:
I've been doing extensive refactoring of TTS-related stuff recently, so I advise against doing anything non-trivial with TTS at the moment.
I may look into the issue myself later.

di72nn:
There seems to be no straightforward solution to this.

Here are some details on how it works now for anyone interested.
1. The app parses article content (HTML) looking for text nodes. Usually those are paragraphs. So a paragraph often is **a unit** of speech that is produced.
2. Then the app sends those units of speech to some TTS engine to synthesize (the TTS engine sends the sound to a sound device by itself). There's no way to **pause** it, only to stop currently produced sound. So pause/resume is not available.
3. If the app needs to pause, it stops currently produced unit and restarts it from the beginning when it needs to resume. As a result it pretty much always "rewinds".

[...]

PR #945, opened 2020-04-01
Refactoring of TTS

di72nn:
A separate PR for TTS. Based on #931.

I tried to improve code and fix potential synchronization issues while keeping the logic mostly intact (except some initialization stages - those had to be changed because TTS didn't work for me).
It is still far from perfect, but that should do for now.

So far I only tested it a bit with Google TTS on an Android 9 emulator and a LineageOS 16 device.

I may make some more improvements, but no promises.

Outstanding issues (in general, not in this PR):
 * [ ] Bad handling of interruptions (and pause/resume in general). Discussed in #939.
 * [x] Reset to the start of the article on interruptions (#939). As @tyndare [mentioned](https://github.com/wallabag/android-app/issues/939#issuecomment-606876811): resume will check that the current text unit is visible on the screen, if it is not the case it will resume at the first text unit visible on the screen.
 * [x] The size of "text units" should be reduced.
 * [x] I'm quite annoyed that text in inline tags is pronounced as separate sentences (e.g. there are pauses around links).
 * [ ] Splitting into sentences is dumb: it will split stuff like `the U.S. and` (it basically splits by any `.`, `?`, `!` followed by a space).

@tyndare here are the latest changes in case you want to give it a go. Please comment if you do.


di72nn:
Inline elements are no longer separate text units.
Text is split into sentences (not perfect, see the first post).

di72nn:
Add rudimentary "emphasizing" for bold/strong/italic text. I didn't find any way to make Android TTS emphasize parts of text, so I just add commas around these emphasized parts to denote them with pauses. I'm not happy with the amount of code (and its complexity) it takes, but oh well.

Also made TTS pronounce alt text or title of images.

di72nn:
Don't reset reading position if scrolling didn't change since pause (fixes #939 as a third option [described here](https://github.com/wallabag/android-app/issues/939#issuecomment-609006540)).

Fixed TTS notification icon: a wallabag logo is now displayed instead of a solid square.
Notification action icons are replaced with better looking ones (without outline).

PR #998, opened 2020-06-10
Suggest all available tags if no filter

makuto:
This makes it much easier to tag articles with tags you've entered before.


di72nn:
Squashed the original commits and then simplified the logic.

di72nn:
Oh, and thanks for improving this functionality @makuto!

di72nn:
Also reworked layout a bit. The element order now as follows:
1. Selected tags.
1. Text field + button.
1. Suggested tags.

My UI skills suck, so you may revert or further improve it.

makuto:
Do I close this now or does someone else merge and close it?

PR #1008, opened 2020-06-25
Add a splash screen

di72nn:
That white startup screen is starting to get to me.

I don't think startup time is going to be improved anytime soon, so here's a splash screen.

To the best of my knowledge it cannot be made light or dark according to in-app theme (because splash screen theme is applied before any app code is executed). So I made it dark, because a dark screen is okay in light environment, but a light screen is annoying in the dark.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -19,16 +19,23 @@
         android:theme="@style/LightTheme"
         android:networkSecurityConfig="@xml/network_security_config">
         <activity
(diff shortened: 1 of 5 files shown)


Strubbl:
Is it possible to get a splash screen when resuming the app? Sometimes ,  when resuming is not that fast, I still get a white fullscreen.

di72nn:
What activity do you resume to?
You can check method 1 vs method 2 [here](https://android.jlelse.eu/the-complete-android-splash-screen-guide-c7db82bce565).
We probably need to create an issue for discussion.

PR #1014, opened 2020-07-13
Make "previous"/"next" buttons honor sort order and filtering

### LINKED ISSUE #553: TTS article reading order
I'm selecting one tag, then ordering the items from older to newer, click in the first article and click in TTS with "read next article" active. First article is read perfectly but the next article is not the 2nd article in the list, is random.

di72nn:
I was going to fix #553, did some refactoring, but got stuck with deciding on behavior: there is some excessive state restoration (like filtering tags, opening article list, erasing filter, going back to tags - the tags will be filtered).

I don't feel like finishing it now.


di72nn:
I recently learned about [Android Paging library](https://developer.android.com/topic/libraries/architecture/paging), which seems to be a good fit to finally implement async lists.
There's a good chance it will be added in this PR.

PR #1019, opened 2020-07-16
Startup optimizations

di72nn:
I'm annoyed with cold startup time when you share to wallabag (the delay between you press the wallabag icon in the share menu and that menu disappearing).

I started with deferring all the init in the `App` class. It shaves off a couple of hundreds milliseconds, but it seems that the bulk of time is actually spent by the system loading the app (should try out resource optimization?).

This PR is ok as it is even though it doesn't yield the desired result.

I'm going to experiment more with this later.


tcitworld:
> I'm annoyed with cold startup time when you share to wallabag (the delay between you press the wallabag icon in the share menu and that menu disappearing).

It looked already quite fine to me (I killed the app before trying) when sharing through the system native share menu or Firefox's new custom thing.

I have Android 10, maybe there was improvements in this area when they introduced Sharing Shortcuts?

Issue #1013, opened 2020-07-12
tagging eventually hangs in 2.4.0beta4


anarcat:
# Issue details

When tagging articles, after a few entries, the apps completely hangs and becomes unresponsive enough for Android to ask me if I want to kill it.

### Duplicate?

no

### Actual behaviour

Fairly reliably, before I get to type "add" on the third tag (see "steps to reproduce", below), the entire app hangs and becomes completely unresponsive. After a while, Android asks me to have mercy on the poor phone's battery and to kill the app.

If I let it be, the Manage tags dialog eventually closes on its own and the app restarts, presumably it crashes. The article stays untagged, both according to the main Wallabag web UI and the Android app.

### Expected behaviour

I should be able to add many tags, and this was indeed working fine in the 2.3.0 version.

# Steps to reproduce the issue

1. Tap on the top-left hamburger menu
2. Tap on the `<untagged>` "tag"
3. Tap on an article
4. Tap on the top-right "three dot" menu
5. Tap on `Manage tags`
6. Type a tag, tap "add"

Repeat the last step two or three times. 

# Environment details
- **wallabag app version**:  2.4.0beta4 (regression from 2.3.0)
- **wallabag app installation source** F-Droid
- **Android OS version**:  7.1.2 AOSP derivative, 2019-10-05 patch level
- **Android ROM**: Fairphone OS ("Saibon")
- **Android hardware**:  Fairphone 2
- **wallabag server version**: 2.3.7
- **Do you have Two-Factor-Authentication enabled?**: no

# Logs
## wallabag server

I do not have access to the server logs.

## Web server

N/A.

# Your experience with wallabag Android app

This app is great! I use it all the time, both on my phone and on my Onyx e-reader, it's pretty awesome. It usually works great and I absolutely *love* the idea of having an "untagged" search (which the Web UI <del>could have too</del> already has! sorry, I forgot! :)), I can't wait to upgrade to the beta everywhere. :)

di72nn:
The report is solid, but I haven't been able to reproduce it so far, so bear with me as I have to ask some clarifying questions.

Is there any background activity (synchronization, article sweeping)?
You don't have to press "save" (the checkmark in the action bar) to reproduce it?
Does it have any relevance whether you type in new tags or already existing ones (existing on other articles)?
Do you add tags one by one or use a comma-separated string with multiple tags?
At which point exactly does the app stop responding (you can try to check responsiveness by scrolling screen / tag lists):
[...]

PR #1022, opened 2020-07-22
Release 2.4.0-beta.5

di72nn:


tcitworld:
Apart from #1024 that I just discovered, we should be good for a stable release if no issues come up, what do you think?

di72nn:
We need to check that #1013 is fixed and it's good to go.
#1024 may be dealt with later.

Issue #633, opened 2017-09-21
Enhancement: convert text-selection to tags


shtrom:
Tags are generally present as words in the relevant articles. Moreover, it's not always obvious from the title of a piece what all the tags for it may be (e.g., a list of tools or techniques).

It would be good to have a quick way to add tags from the text, without having to Copy/Manage tags/Paste/Add/Save.

The less intrusive I can think of would be to add an `Use as tag‘, or similar, context menu option on text selections.

PR #1053, opened 2020-09-07
Add context menu to add tag from selection

shtrom:
fixes: #633
Signed-off-by: Olivier Mehani <shtrom@ssji.net>

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ManageArticleTagsActivity.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ManageArticleTagsActivity.java
@@ -88,16 +88,13 @@ public class ManageArticleTagsActivity extends BaseActionBarActivity {
         setContentView(R.layout.activity_manage_article_tags);
 
         String[] currentTagsArray = null;
-
-        Intent intent = getIntent();
-        String text = intent.getStringExtra(PARAM_TAG_LABEL);
-        if (
-                text == null
-                && savedInstanceState != null
-            ) {
+        String text;
+        if (savedInstanceState != null) {
             discoveredArticleId = savedInstanceState.getInt(STATE_DISCOVERED_ARTICLE_ID, -1);
             currentTagsArray = savedInstanceState.getStringArray(STATE_CURRENT_TAGS);
             text = savedInstanceState.getString(STATE_CURRENT_TEXT);
+        } else {
+            text = getIntent().getStringExtra(PARAM_TAG_LABEL);
         }
 
         if (!loadArticle()) return;
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ManageArticleTagsActivity.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ManageArticleTagsActivity.java
@@ -52,6 +52,7 @@ public class ManageArticleTagsActivity extends BaseActionBarActivity {
 
     public static final String PARAM_ARTICLE_ID = "article_id";
     public static final String PARAM_ARTICLE_URL = "article_url";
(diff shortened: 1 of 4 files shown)


di72nn:
Thanks for the PR and sorry for the delay!

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
```diff
@@ -398,22 +398,41 @@ public boolean dispatchTouchEvent(MotionEvent ev) {
...
+        } else {
+            annotateItem.setVisible(false);
         }
 
         super.onActionModeStarted(mode);
     }
 
+    private void createTagFromSelection(String selection) {
+        selection = selection.replaceAll("^\"|\"$", "")
+                .replaceAll("\\\\n|\\\\r|\\\\t", " ");
```
di72nn, line 428: Are you sure this does what you meant it to do?

di72nn:
I haven't run it yet.

To merger: squash.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
shtrom: I got to that by trial and error. Essentially, if you select some text with newlines and/or tabs, they get encoded in the string, but seem to be escaped. So we need to find and replace the escaped whitespaces...

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ManageArticleTagsActivity.java
```diff
@@ -88,16 +88,13 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
-                text == null
-                && savedInstanceState != null
-            ) {
+        String text;
+        if (savedInstanceState != null) {
             discoveredArticleId = savedInstanceState.getInt(STATE_DISCOVERED_ARTICLE_ID, -1);
             currentTagsArray = savedInstanceState.getStringArray(STATE_CURRENT_TAGS);
             text = savedInstanceState.getString(STATE_CURRENT_TEXT);
+        } else {
+            text = getIntent().getStringExtra(PARAM_TAG_LABEL);
```
shtrom: I'm not exactly sure what `savedInstanceState` is, but this seems to change the precedence. I assumed that  `savedInstanceState` was, maybe, a previously entered tag, but not added, before switching to another app.

In this case, and if my assumption is correct, I figured that it might make more sense to ignore the previously entered tag, and show the new selection instead, which is the purpose of the intent.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
di72nn: Yup, works correctly.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ManageArticleTagsActivity.java
di72nn: `savedInstanceState` is only present when the activity is recreated (like when screen orientation changes), it contains saved state which may be newer (you already edited the string) than the original string in the intent.

That may be complicated with non-standard activity parameters (affinities and stuff), but we don't have anything like that at the moment.

di72nn (approved):
Looks good. Tested on Android 9.

To merger: please squash.

tcitworld:
Squashed and merged in https://github.com/wallabag/android-app/commit/a8e0888396df4f65042a766fe982acb6560401e0

Issue #974, opened 2020-05-13
Foreground services


di72nn:
Normal non-foreground [services](https://developer.android.com/guide/components/services) are somewhat downgraded in modern Android versions: there are some background restrictions starting from Android 8, the services can't be started from background (#784), they can be killed by the system.

#931 replaces the service implementation, but still uses the same non-foreground services.

Here are some thoughts on how this should be fixed.

One of the main obstacles to simply replacing non-foreground services with foreground services is the requirement to display a notification. Not only it would be annoying for the user (to see a notification pop up even for short actions), but it also requires to rewrite the way notifications are displayed in the wallabag app.
Another reason against dumb transition to foreground services is the reason that not everything should be handled by foreground services.

In accordance with the docs, foreground services should handle long running *immediate* tasks.
I think in case of the app the only task that qualifies is the synchronization. Maybe also image fetching (or maybe it can be done with a WorkManager job).

Other short actions (like article modifications) should still be handled by background services, which should do fine under normal conditions. If such tasks are blocked by some other background operation (sync), the background service executing them should promote itself to a foreground service (in order to not be killed and lose data).

Some of the tasks (like background sync triggered by schedule or network events) should be done using WorkManager jobs. Now that the task logic is moved out of services it shouldn't be too hard.

For a while in the app the tasks haven't been not explicitly synchronized with each other. The synchronization relies on the fact, that the tasks are executed sequentially by a single service.
Introducing multiple task executors (foreground, background, WorkManager) requires to explicitly synchronize task execution. Maybe an exclusive SQLite write transaction is enough, maybe explicit locks are a better choice, I haven't look into it much yet.

As usual, these are notes for anyone willing to do some development.

shtrom:
I think I ran into this issue when setting up a new client, and the app would never succeed to do the initial sync and fail with a `java.io.InterruptedIOException`

![Screenshot_20200923-193335_Phone](https://user-images.githubusercontent.com/160280/94683922-9d92ea80-036a-11eb-86f2-a434a7f7b719.png)

As a workaround, disabling battery optimisations for Wallabag worked for this initial sync, and it seemed to be better able to do the rest of the sync later on.

![Screenshot_20200923-193437_Settings](https://user-images.githubusercontent.com/160280/94683863-88b65700-036a-11eb-9ffb-5d898d6b22e1.png)

shtrom:
[...]

Issue #784, opened 2019-05-04
Google Play issue : java.lang.IllegalStateException: 


tcitworld:
```
java.lang.IllegalStateException: 
  at android.app.ContextImpl.startServiceCommon (ContextImpl.java:1521)
  at android.app.ContextImpl.startService (ContextImpl.java:1477)
  at android.content.ContextWrapper.startService (ContextWrapper.java:650)
  at fr.gaulupeau.apps.Poche.service.ServiceHelper.startService (ServiceHelper.java:221)
  at fr.gaulupeau.apps.Poche.service.ServiceHelper.startService (ServiceHelper.java:204)
  at fr.gaulupeau.apps.Poche.service.ServiceHelper.syncQueue (ServiceHelper.java:103)
  at fr.gaulupeau.apps.Poche.service.ServiceHelper.syncQueue (ServiceHelper.java:97)
  at fr.gaulupeau.apps.Poche.events.EventProcessor.networkChanged (EventProcessor.java:579)
  at fr.gaulupeau.apps.Poche.events.EventProcessor.access$000 (EventProcessor.java:37)
  at fr.gaulupeau.apps.Poche.events.EventProcessor$1.run (EventProcessor.java:586)
  at android.os.Handler.handleCallback (Handler.java:790)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loop (Looper.java:164)
  at android.app.ActivityThread.main (ActivityThread.java:6518)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:438)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:807)
```

Happening on `Asus ZenFone Max Pro (M1) (ASUS_X00T_6), Android 8.1`, version 210.

tcitworld:
Another one seems the same:
```

java.lang.RuntimeException: 
  at android.app.ActivityThread.performResumeActivity (ActivityThread.java:3987)
  at android.app.ActivityThread.handleResumeActivity (ActivityThread.java:4019)
  at android.app.servertransaction.ResumeActivityItem.execute (ResumeActivityItem.java:51)
  at android.app.servertransaction.TransactionExecutor.executeLifecycleState (TransactionExecutor.java:145)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:70)
[...]

PR #1068, opened 2020-10-01
Foreground services

### LINKED ISSUE #975: Auto-sync is broken
As far as I know, both "Enable auto-sync" (scheduled background synchronization) and "Separate auto-sync for local changes" (background sync triggered by network availability) [are broken](https://github.com/wallabag/android-app/issues/784) on Android ~8+.

The problem is that the background services we use can't be started while the app is in the background. The situation is explained to some extent in #974.

di72nn:
I'm not sure if I'll be able to properly deal with #974 anytime soon, so here's a not so elegant but simple fix.

This PR should fix a whole slew of `IllegalStateException`s reported in the Play Console. Fixes #975, #784.

Using foreground services has some merit for some of the tasks, but #974 still stands.

First commit here makes the app use foreground services for any background tasks. This is technically ok, but Android requires to display a visible notification when the service is running, which is quite annoying (especially because in this implementation the notifications don't have any meaningful content). These notifications may be disabled (using notification channels), but *only manually by the user*.

According to Play Console the errors are relatively rare, so the app should work fine with non-foreground services *most of the time*. The second commit tries to use foreground services only when starting a normal service fails. This is debatable.

The simplest way to test it is to enable autosync on app start, and start the app with the screen turned off (like with `adb shell am start -n fr.gaulupeau.apps.InThePoche.debug/fr.gaulupeau.apps.Poche.ui.MainActivity` or directly from AndroidStudio). There's a timeout in `MainActivity.updateOnStartup()` though.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/service/MainService.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/service/MainService.java
@@ -1,9 +1,32 @@
 package fr.gaulupeau.apps.Poche.service;
 
+import android.app.Notification;
+
+import androidx.core.app.NotificationCompat;
+
+import fr.gaulupeau.apps.InThePoche.R;
+import fr.gaulupeau.apps.Poche.ui.NotificationsHelper;
+
 public class MainService extends TaskService {
 
     public MainService() {
         super(MainService.class.getSimpleName());
+    }
+
+    @Override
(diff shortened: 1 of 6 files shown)


tcitworld (approved):

tcitworld:
Thanks for taking time to explain the details, btw!

Strubbl (approved):

PR #1085, opened 2020-10-16
update gradle wrapper to v6.5

Strubbl:


di72nn:
1. Whenever you change gradle wrapper version, run `gradle wrapper --gradle-version 6.5 --distribution-type all` to update the scripts.
2. It's not the case this time, but in general be warned that F-Droid may not build projects with recent gradle versions.
  Even though [their sources have info about `6.7`](https://gitlab.com/fdroid/fdroidserver/-/blame/master/gradlew-fdroid#L144), they don't use the latest builds in production (`buildserverid` in [some recent build](https://f-droid.org/wiki/page/org.gateshipone.odyssey/lastbuild_32) is more than a month old, so `6.6.1` is the latest version that would currently build).

di72nn (approved):

PR #1107, opened 2020-12-01
Android 4 compatibility

di72nn:
I wanted to know which things are compatible with Android 4 and how hard it is to maintain compatibility.
Not worth it. So the remaining legacy version checks should be removed from master.

Tested on Android 4.0.3 emulator (API 15).

There's a security issue with JavaScript having access to any Java methods (not only specifically annotated to work with JS) on the injected into WebView objects on Android <4.2.

WebView: `window.getSelection()` seems to not work (maybe an emulator issue, though). Maybe workarounds are possible, haven't checked it. As a result:
* "Add tag" from selection doesn't work.
* Annotation creation doesn't work. Displaying, editing and deleting annotations works (to my surprise).

Full-text-search is enabled only since Android 4.1. Android 4.0 uses the old title-based search.

TTS: parsing seems to work, didn't check the work with TTS engines. Guaranteed degraded fast-forward/rewind functionality on Android <4.0.3.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -72,6 +72,6 @@ dependencies {
     implementation 'com.facebook.stetho:stetho:1.5.1'
     implementation 'com.facebook.stetho:stetho-okhttp3:1.5.1'
     implementation 'com.mikepenz:aboutlibraries:6.2.1'
-    implementation 'com.github.di72nn.wallabag-api-wrapper:api-wrapper:v2.0.0-beta.5'
+    implementation 'com.github.di72nn.wallabag-api-wrapper:api-wrapper:d7a502ac7b'
     implementation 'org.slf4j:slf4j-android:1.7.30'
 }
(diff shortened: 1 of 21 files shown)


di72nn:
Rebased on the latest release.

Linter fails and suggests to add `android.defaultConfig.vectorDrawables.useSupportLibrary = true` which satisfies it, but **makes the app crash at runtime** on Android 4.
As a result, `./gradlew build` fails, but `./gradlew assemble` assembles APKs (without performing checks) just fine.

tristan-k:
I'm really looking forward to Android 4 compatibility because the Tolino Shine 3 only runs on Android 4.4.2.

di72nn:
@tristan-k there's nothing to look forward to - this PR is not going to be merged. You can build the app from this branch yourself - just use `gradle assemble`.

PR #1111, opened 2020-12-06
Replace Travis CI with GitHub Actions

di72nn:
I'm not a fan of GitHub, but I'm ok with using its CI as long as we don't **depend** on it.
Hopefully, it won't be [as flaky as Travis](https://travis-ci.org/github/wallabag/android-app/builds/742903404) for Android builds.


di72nn:
Also disabled gradle wrapper validation for non-master branches (to avoid duplicate validations for `push` and `pull_request`).

tcitworld (approved):

tcitworld:
Quite faster as well.

di72nn:
I can't figure out why that `$default-branch` macro doesn't work (there's no check for the latest commit in `master`). It was copied from [the official example](https://github.com/actions/starter-workflows/blob/main/ci/android.yml).

I guess I'll just hardcode `master` if it doesn't magically get fixed by itself.

PR #1112, opened 2020-12-07
CI: hardcode default branch

di72nn:
I'm not 100% sure it's the reason, but that's my best guess.


tcitworld (approved):

di72nn:
Works now.
Good job, github. /s

PR #1143, opened 2021-03-30
Updated quote style

### LINKED ISSUE #1141: Updated quote style
Hi, I have some draft changes to the quote style making them closer to the style on the Wallabag web service.

eg. From ...

image

To ...

image

Is that of interest? I can send through a PR. `pre` blocks are unchanged.

creywood:
Edit: Now better integrated at #1165

~~As discussed at #1141, the updated style aims to be more easily identified as a quote and is more closely matched to the Wallabag web service style.~~

~~Closes #1141.~~

### DIFF
--- a/app/src/main/assets/dark.css
+++ b/app/src/main/assets/dark.css
@@ -74,7 +74,8 @@ img {
 }
 
 .high-contrast p,
-.high-contrast li {
+.high-contrast li,
+.high-contrast #article blockquote {
   color: #FFF;
   background: #000;
 }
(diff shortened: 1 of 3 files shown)


di72nn:
Sorry for the delay and thank you for the PR!

Any idea why the height of the area marked with red is about twice the size of the area marked with green? Seems like the `p`'s `margin-bottom` causes it, but I'm not very experienced in CSS to tell for sure.

quote

Also, I'm testing your commit rebased on master, but I don't think it matters.

BTW, if you didn't know, you can use `chrome://inspect` in chromiun-based browsers to remotely inspect the WebView content of debug builds of the app.

di72nn:
Compared with wallabag's web UI - seems like it has something to do with "collapsing margins" not collapsing. Not something I know how to handle.

di72nn:
No, you're right, the problem is present in our current style too. We can merge this and fix the problem separately.

di72nn (approved):
Seems fine, but CSS is not my strength.

The CI status doesn't matter - the job was ran against master when it has an issue (#1142).

di72nn:
Do not merge. Probably replaced with #1165.

Strubbl:
I think we can close this in favor of #1165. Or did i overlook something this PR fulfills, which the other doesnt?

PR #1159, opened 2021-05-06
Update gradle

di72nn:
Update Android gradle plugin and gradle wrapper (requirement for the plugin).


tcitworld (approved):
The graddle-wrapper.jar has been removed, is that intended? In that case, is the `gradle/wrapper-validation-action` still needed?

di72nn:
> The graddle-wrapper.jar has been removed, is that intended?

No, it's there, it just isn't obvious from the diff.

tcitworld:
Indeed, I was tricked by Github's mobile app.

PR #1161, opened 2021-05-06
Use creation date instead of article ID for sorting

di72nn:
I'll need to benchmark it, probably add an index.

Either I miss something or it's impossible to get "previous/next article" using SQL if creation dates may be duplicate **and** grow in a different direction from IDs. It is possible to select multiple articles and choose one on the Java side, but I don't think it's worth it.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/service/workers/ArticleImagesFetcher.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/service/workers/ArticleImagesFetcher.java
@@ -66,7 +66,7 @@ public class ArticleImagesFetcher extends BaseNetworkWorker {
         QueryBuilder<Article> queryBuilder = articleDao.queryBuilder()
                 .where(ArticleDao.Properties.ArticleId.isNotNull())
                 .where(ArticleDao.Properties.ImagesDownloaded.eq(false))
-                .orderAsc(ArticleDao.Properties.ArticleId);
+                .orderAsc(ArticleDao.Properties.CreationDate, ArticleDao.Properties.ArticleId);
 
         int totalNumber = (int) queryBuilder.count();
         Log.d(TAG, "fetchImages() total number: " + totalNumber);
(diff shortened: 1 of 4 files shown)


elzik:
It would be nice to see this merged. I have just bumped into #713 after importing from Omnivore and I had to rejig the ID of all the entries to get the Android app to display things in a more useful order.

jason-mehmel:
@elzik How did you rejig the ID of all of the entries?

I ran into the exact same issue.

elzik:
> @elzik How did you rejig the ID of all of the entries?

I connected to the SQLite database directly and changed all the IDs of the entries as necessary. You would have to adjust the foreign keys to match.

Pretty dirty, but until this PR is merged, I am not sure what the other options would be.

Janneaa:
I really hope that this change is implemented soon. Now that Pocket is closing I'm sure that there will be lots of people importing large amounts of articles from Pocket to Wallabag. I've also imported thousands of them and having them sorted almost randomly in the Android app is really frustrating.

tahr35:
For me, it is logical and understandable that the IDs change after an import.
However, the display (sorting) in the app should be corrected. 
Just like in the web version according to the **creation date**.

I also hope for an update soon. 😀
Maybe the developer can also add a selection option for the sorting method if he is not sure about such a profound change. Although for me, sorting by creation date like in the web version is actually the most logical and manageable.

webformation:
The android application is quite useless when we do an import of pocket with several files. 

I found no way to do the import in order to have articleID in the order of date.

Merging this PR would be a good improvement.

tahr35:
The article ID cannot be adjusted, it must be reassigned, otherwise duplicates may occur. 
This makes sense and is logical. 
Only the sorting of the list (in the app) is the problem.

Sorting should be based on the **creaction date**, as in the web application, not on the article ID.

Strubbl:
I rebased this branch on current master branch and built a debug version if anyone is fancy to try this out:

[app-debug.apk.zip](https://github.com/user-attachments/files/20579364/app-debug.apk.zip)

You can install the debug version in parallel to the release version you installed.

Strubbl:
When this PR is going to be merged without addressing the know issues, I created #1456 and #1457 as regression tickets for a later follow-up.

Strubbl (approved):
I also did test these changes. No finding. Not even performance problems. Works very well although i did not see a difference because my ID numbers and dates grew over a long time.

PR #1165, opened 2021-05-12
Update quote style and remove bottom margin from paragraphs within them

### LINKED ISSUE #1141: Updated quote style

### LINKED ISSUE #1163: Quotes have some excessive space at the bottom
Discussed in #1143.

quote

creywood:
@di72nn as you identified this was to do with padding on the bottom of paragraphs within the blockquote. Looks good for me for articles which were and weren't affected by the issue.

Fixes #1163
Closes #1141

### DIFF
--- a/app/src/main/assets/solarized.css
+++ b/app/src/main/assets/solarized.css
@@ -235,6 +235,10 @@ header.mbm {
   margin: 0;
 }
 
+#article blockquote p {
+  margin-bottom: 0;
+}
+
 #article pre {
   font-family: monospace;
   white-space: pre;
--- a/app/src/main/assets/solarized.css
+++ b/app/src/main/assets/solarized.css
@@ -232,7 +232,7 @@ header.mbm {
   border:1px solid #999;
   background: #eee8d5; /* $base2*/
   padding: 0.25em;
-  margin: 0;
+  margin: 1.25em 0;
 }
 
 #article blockquote p:last-of-type {
--- a/app/src/main/assets/solarized.css
+++ b/app/src/main/assets/solarized.css
@@ -229,17 +229,21 @@ header.mbm {
 }
 
 #article blockquote, #article pre {
-  border:1px solid #999;
-  background: #eee8d5; /* $base2*/
-  padding: 0.25em;
   margin: 1.25em 0;
 }
 
-#article blockquote p:last-of-type {
-  margin-bottom: 0;
+#article blockquote {
+  border-left: 5px solid #b58900; /* $yellow; */
+  padding-left: 1.25rem;
+  padding-right: 0.25rem;
+  font-style: italic;
+  color: #657B83; /* $base00 */
 }
 
 #article pre {
+  padding: 0.25em;
+  background: #eee8d5; /* $base2*/
+  border: 1px solid #999;
   font-family: monospace;
   white-space: pre;
   text-justify: none;
@@ -253,7 +257,7 @@ header.mbm {
   white-space: pre-wrap;
 }
 
-#article blockquote *, #article pre * {
+#article pre *  {
   background: inherit;
 }
 
--- a/app/src/main/assets/solarized.css
+++ b/app/src/main/assets/solarized.css
@@ -78,7 +78,8 @@ table, th, td {
 }
 
 .high-contrast p,
-.high-contrast li {
+.high-contrast li,
+.high-contrast #article blockquote {
   color: #000;
   background: #FFF;
 }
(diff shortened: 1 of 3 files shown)


di72nn:
Thanks!

I haven't tested it, but I believe it will also remove margins between multiple paragraphs in a quote. Have you checked that?

creywood:
I was wondering about that too (... and to be honest I think it then slipped my mind :facepalm: ). They also don't often occur in my feeds.

Do you happen to have a feed/article I could test this with?

creywood:
I've just put up a fix, but have thus far only tested it with single paragraph quotes (which it works for)

di72nn:
Multiple `p`'s in a quote: https://zerforschung.org/posts/gorillas-en/
Many different quotes, some adjacent quotes are visually merged: https://towardsdatascience.com/5-deep-learning-trends-leading-artificial-intelligence-to-the-next-stage-11f2ef60f97e
`blockquote p em`: https://www.theverge.com/2021/5/11/22430185/alexa-voice-actor-amazon-nina-rolle
`blockquote a` that is displayed fine: https://blog.cryptographyengineering.com/2021/03/25/whats-in-your-browser-backup/

creywood:
Hrmm yeah the rest look ok but the towards data science article shows another problem which would be exacerbated if these PRs are merged. It looks like sometimes there's no spacing left after the `blockquote`. Time permitting I'll have a look.

creywood:
I've had a go at that too. Blockquotes will be spaced out a little more, closer to wallabag web, which looks fine to me but may be subjective.

di72nn:
Thanks for your efforts!

The changes seem to cover most of the problems, except for `p` being not the last element in the quote (the blog.cryptographyengineering.com example).

However, I feel like the web UI does it somehow differently and trying to borrow that approach is a better way to go.

creywood:
I agree, having more coverage ala the wallabag web stylesheet would avoid these issues, with the benefit of (arguably) improved style (including eg. the improved quote style that kicked this off). I think effectively tracking the web stylesheet is a good idea.

I can remove these PRs if you'd like.

di72nn:
> I can remove these PRs if you'd like.

No, I don't think anything major is going to happen to styles in the app (at least anytime soon), so these PRs still make sense.

But for this fix we should try collapsing bottom margins (like in the web UI, if I understand correctly) instead of removing margins from `p`.

creywood:
Sure thing. Have made changes to allow for margin collapsing. As a result integrated #1143 changes as that helps in removing the border between the parent `blockquote` and its children (which was part of what was preventing the collapsing).

FYI won't be back in internet reception for half a week or so.

creywood:
By the way, sorry for the rushed approach to my changes on the margin issue, FWIW that's not normally my way

di72nn:
> By the way, sorry for the rushed approach...

Don't worry, it's ok :-)

Looks nice, thanks again!

Can you please double-check the `high-contrast`-related part. It was changed in `main.css`, but not in `dark.css` and `solarized.css` (I know we don't use high-contrast with solarized, but I'd prefer to have it for consistency, if it doesn't contradict anything).

app/src/main/assets/solarized.css
```diff
@@ -249,7 +257,7 @@ header.mbm {
   white-space: pre-wrap;
 }
 
-#article blockquote *, #article pre * {
+#article pre *  {
```
di72nn, line 260: Can you please remove the extra space, it's driving me nuts :-D

creywood: Oops, sorry, see comment re. rushing. Done.

creywood:
> > By the way, sorry for the rushed approach...
> 
> Don't worry, it's ok :-)
> 
> Looks nice, thanks again!
> 
> Can you please double-check the `high-contrast`-related part. It was changed in `main.css`, but not in `dark.css` and `solarized.css` (I know we don't use high-contrast with solarized, but I'd prefer to have it for consistency, if it doesn't contradict anything).

Yeah, I think I'd excluded it from solarized as it wasn't used, and for dark who knows why I didn't. The more contrast the better, right? :slightly_smiling_face:  Updated.

di72nn (approved):
Squash on merge.

di72nn:
Everything looks good, thanks again!

Strubbl (approved):

Issue #742, opened 2018-11-03
[feature request] create public links


nr458h:
It would be nice to be able to create public links similar to the web interface.

nr458h:
In the articles setting is one choise *Copy original URL* and there I would expect also a choise like *Create Public URL* or something similar.

Strubbl:
And after selection that menu option? 

Shall the public link URL be available in the app somehow or does it not matter what the public URL is? Or just a toast message with or without the link URL?

nr458h:
I would expect the same behavior as the copy original URL function has: that the URL is now copied to the clipboard. That would fit my specific usecase.

But in my opinion there could be in addition also the option to choose would URL gets shared when the share button is selected

biva:
Hello, any news about this feature? It would be great!

To make it easy, I would just add a button in the menu, below the existing "Share" button. The list of actions would be:
Share original link 
Share public link 
Modify the title
... 
Thanks in advance!

retiolus:
Any news about this feature? For me, this is one of the most important things that is missing... I have to open the web interface every time.

Strubbl:
I am not sure if the API does expose a method to implement this feature. Am i overlooking anything? https://app.wallabag.it/api/doc

di72nn:
There's a `public` flag with description "will generate a public link for the entry" on [`
PATCH /api/entries/{entry}`](https://app.wallabag.it/api/doc#patch--api-entries-{entry}.{_format}) and [there's a field](https://github.com/di72nn/wallabag-api-wrapper/blob/5570ac6aa51ee41492a991874bebeb4af9b198d3/api-wrapper/src/main/java/wallabag/apiwrapper/models/Article.java#L115) allowing to construct a public link.
Shouldn't be a lot of work, mostly implementing UI, but I haven't gotten around to doing it.

retiolus:
I tried to add this option but so far it only works when the article has already been made public once through the web interface.

```
case R.id.menuCopyPublicURL:
                article.setIsPublic(Boolean.TRUE);
                copyUrlToClipboard(activity, url + "/share/" + article.getPublicUid());
                return true;
``` 
I don't quite understand what `article.setIsPublic(Boolean.TRUE);` does, since it doesn't make the article public.

![image](https://user-images.githubusercontent.com/65604466/199379275-5c7b3a2a-51b1-4ab3-8703-16ada9f3dd81.png)

matt-wirth:
[...]

PR #1285, opened 2022-11-02
Support for creating and sharing public url from the app

retiolus:
#742


di72nn:
I guess I understated the complexity of this for a newcomer. Actually, even I spent an hour remembering how all this works and I'm still having trouble putting it all together :)

I'll probably have to implement it myself (because the codebase is more complex than it should have been), but here's some pointers so you can hack something usable for yourself.

The very minimum that needs to be done is a call like this:
```java
        Article article = wallabagService.modifyArticleBuilder(articleId)
                .isPublic(true)
                .execute();
```
That would request the server to "share" the article. The returned `article` would contain the `publicUid` field.

The problem is that Android would only allow you to do that in a non-UI thread (because it's a "long" network operation). I guess the simplest way would be to use `AsyncTask`.

I'll try to implement it properly in the near future, but can't promise anything.

PR #1393, opened 2024-04-13
fix translation after PR #1268

Strubbl:
after this PR is merged, translation via weblate should be possible again

PR #1394, opened 2024-04-13
fix deprecation warning for stopForeground()

Strubbl:


Strubbl:
This build fails due to #1393 being a dependency.

tcitworld (approved):

PR #1406, opened 2024-05-22
Remove non-foss greendao plugin

### LINKED ISSUE #1399: F-Droid issue: The greendao gradle plugin is not FOSS
# Issue details

Since https://github.com/wallabag/android-app/commit/f8493cc65fe566326ae557780b1ef63da72adf1d the greendao generator is replaced with the gradle plugin. But the source code of the plugin is never released, see https://github.com/greenrobot/greenDAO/issues/412. So F-Droid has to remove all affected versions of wallabag and can't build new versions. Is that possible to migrate to a different FOSS lib, since greendao is not maintained anymore? Please note that objectbox is not FOSS yet, see https://github.com/objectbox/objectbox-java/issues/1102.

di72nn:
Add generated DAO classes, remove the generator plugin. I just added the generated classes as is.

A temporary measure to allow app updates be published on F-Droid. The proper solution would be to get rid of greendao, but I can't look into it at the moment.

Kinda resolves #1399.

Upd: clarification: without the generator the DAO classes are not automatically updated when the model classes change (but the model hasn't changed in a while).


app/src/main/java/fr/gaulupeau/apps/Poche/data/dao/AnnotationDao.java
```diff
@@ -0,0 +1,243 @@
...
+import org.greenrobot.greendao.Property;
+import org.greenrobot.greendao.internal.DaoConfig;
+import org.greenrobot.greendao.database.Database;
+import org.greenrobot.greendao.database.DatabaseStatement;
+import org.greenrobot.greendao.query.Query;
+import org.greenrobot.greendao.query.QueryBuilder;
+
+import fr.gaulupeau.apps.Poche.data.dao.entities.Annotation;
+
+// THIS CODE IS GENERATED BY greenDAO, DO NOT EDIT.
```
Strubbl, line 17: the 8 occurances of this comment shall be removed according to https://github.com/wallabag/android-app/issues/1399#issuecomment-2127209508

Strubbl (approved):

blundell:
@di72nn I'm looking to do the same on another project. Could you briefly explain where the generated classes are located? and any other info about this migration that could help pls :-)

di72nn:
Sure, @blundell. This is not a proper solution, mind you.

First, you have to build the project locally at least once, so the generator plugin actually generates the source files.

Then the generated sources should be in `app/build/generated/source/greendao/<your package>`, where `<your package>` is the DAO package (defined by `daoPackage` in one of the `build.gradle` files).

In a non-complicated scenario this is just a matter of moving files to the main source tree (you should be able to drag-n-drop, maybe creating the target package in advance):
Screenshot_20240527_220131

After that the greendao plugin can be removed from `build.gradle` files (see the PR changes).

blundell:
Awesome, appreciate that - going to attempt it this week. TY!

PR #1434, opened 2024-10-02
update to gradle 8.7 and wrapper to 8.9

Strubbl:


Strubbl:
Strange, i only restarted the CI with enabled debug logging and then it passed. Also i couldn't reproduce the former issuer on my machine.

tcitworld (approved):

PR #1482, opened 2025-09-19
Fix: http schema handling

KevinSJ:
The fr.gaulupeau.apps.Poche.ui.HttpSchemeHandlerActivity is supposed to be able to handle any http or https url. 

However, as part of the change in #1434, this functionality was broken, it is not no longer possible to set wallabag to handle http scheme even when the setting option is set to true.

Reverting the change fixes this issue.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -89,7 +89,6 @@
 
                 <data android:scheme="http" />
                 <data android:scheme="https" />
-                <data android:host="app.wallabag.it" />
             </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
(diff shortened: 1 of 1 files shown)


Strubbl:
What do you mean with "handle http scheme"? I do not get what is fixed with this commit.

KevinSJ:
@Strubbl So there is an option in wallabag settings to handle http scheme, which essentially makes wallabag appear as a browser. The PR I mentioned break this functionality.
Screenshot_20260428_093222_wallabag

Strubbl:
Thanks for explaining. I need to check if we still need this at all and/or if it is mandatory for sharing URLs to wallabag app.

KevinSJ:
I don't think it's mandatory, but it does make things a lot more convinient. If we no longer need this, we should delete the option from settings.

Strubbl:
Which convenience? Convenience sounds like a reason to keep the setting and make it work again.

PR #1486, opened 2025-11-24
opt-in for ARM memory tagging extension

Strubbl:


Strubbl:
Sorry, i did miss to add some explanation why i want this change. This extension helps if we activate it:

> Armv9 introduced the Arm Memory Tagging Extension (MTE), a hardware extension that allows you to catch use-after-free and buffer-overflow bugs in your native code.

source and more info: https://developer.android.com/ndk/guides/arm-mte#gradle

Issue #616, opened 2017-07-19
Add image zoom


pVesian:
Hi,

It would be great to be able to:

- tap an image, it would open in fullscreen mode
- be able to (un)zoom in that fullscreen mode

Thanks!

brabalan:
I wanted to file an issue for this too, but I see that it is here. I'm interested in this feature as well.

Valdaer:
Me too, wallabag is great, I've been using it for more than 2 years already but I miss this particular feature from Pocket.

jtagcat:
Saving / sharing would also be nice. With images online-only, it's possible by opening the link in a browser (and zoom there). With offline images, it is not.

PR #1491, opened 2026-04-17
Tapping on an image opens pinch-to-zoom view

amarek:
Some images embedded in articles become too small to see the details when viewed on a mobile/e-reader device. This change adds the capbility to open a zoomable view of an image by tapping on it. Tapping again closes the view.

Details:
 - Adds an image-zoom handler (JS) that sends image URLs from the article WebView back to the app.
 - New ImageViewActivity (PhotoView-based) opens a full-screen pinch-to-zoom viewer, loading from cache, file://, or network with downsample-to-4096px to avoid Canvas size crashes.
 - Wires the JS interface and activity into ReadArticleActivity and the manifest.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -56,6 +56,9 @@
             android:hardwareAccelerated="true"
             android:configChanges="keyboardHidden|orientation|screenSize"/>
         <activity
+            android:name="fr.gaulupeau.apps.Poche.ui.ImageViewActivity"
+            android:theme="@style/Theme.AppCompat.NoActionBar" />
+        <activity
             android:name="fr.gaulupeau.apps.Poche.ui.ManageArticleTagsActivity"
             android:label="@string/manageTags_title" />
         <activity
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -57,6 +57,7 @@
             android:configChanges="keyboardHidden|orientation|screenSize"/>
         <activity
             android:name="fr.gaulupeau.apps.Poche.ui.ImageViewActivity"
+            android:configChanges="keyboardHidden|orientation|screenSize"
             android:theme="@style/Theme.AppCompat.NoActionBar" />
         <activity
             android:name="fr.gaulupeau.apps.Poche.ui.ManageArticleTagsActivity"
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ImageViewActivity.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ImageViewActivity.java
@@ -1,0 +1,136 @@
+package fr.gaulupeau.apps.Poche.ui;
+
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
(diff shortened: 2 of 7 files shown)


Strubbl:
Thank you, this would have fixed #616. But i do not want to add a dependency to an unmaintained lib (https://github.com/Baseflow/PhotoView). Last commit was 6 years ago, last tag/release ~8 years ago.

Strubbl (changes requested):
as stated above, if we introduce a new dependency it should at least be maintained

amarek:
Replaced PhotoView with https://github.com/panpf/zoomimage which seems to be maintained.

app/src/main/res/values-in
Strubbl, line 1: I hope this file did not serve any purpose.

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ImageViewActivity.java
```diff
@@ -0,0 +1,136 @@
...
+                if (bitmap != null) return bitmap;
+            } catch (Exception e) {
+                Log.w(TAG, "loadBitmap() file URL load failed", e);
+            }
+        }
+
+        // Fall back to loading from network
+        try {
+            URL url = new URL(imageUrl);
+            try (InputStream is = url.openStream()) {
```
Strubbl: Can you change this to use the OkhttpClient here? WallabagConnection.createClient() could be reused (automatically reusing the app's global network settings).

app/src/main/java/fr/gaulupeau/apps/Poche/ui/ImageViewActivity.java
```diff
@@ -0,0 +1,136 @@
...
+                return decodeBytesScaled(buffer.toByteArray());
+            }
+        } catch (Exception e) {
+            Log.w(TAG, "loadBitmap() remote load failed", e);
+        }
+
+        return null;
+    }
+
+    // Canvas hardware-accelerated draw limit is ~100MB; cap at 4096px (64MB @ ARGB_8888).
```
Strubbl: 64 MB is a lot, maybe we need to catch any OOM errors when we use this high resolution? This might be an issue esp. with low-end devices.

app/src/main/AndroidManifest.xml
```diff
@@ -55,6 +55,9 @@
             android:name="fr.gaulupeau.apps.Poche.ui.ReadArticleActivity"
             android:hardwareAccelerated="true"
             android:configChanges="keyboardHidden|orientation|screenSize"/>
+        <activity
+            android:name="fr.gaulupeau.apps.Poche.ui.ImageViewActivity"
+            android:theme="@style/Theme.AppCompat.NoActionBar" />
```
Strubbl, line 61: We should add the `android:configChanges="keyboardHidden|orientation|screenSize"` here, analog to the other activities, so that this image view acitvity does not get recreated when we rotate the phone while viewing the image.

Strubbl (changes requested):
Thank you for the update of the dependency. What do you think about my review comments?

amarek:
Addressed the review comments in the latest commit.

PR #1492, opened 2026-04-19
Add customisable automatic dark/light theme switching

MarkusLitz:
## Summary
This PR adds the ability for the app to automatically switch between dark and light themes based on the system's dark mode settings. It also allows the user to select which specific light and dark themes they prefer to use for this automatic switching.

## Features Added
- **Auto dark mode toggle**: A new setting to enable/disable automatic theme switching based on system configuration.
- **Customizable auto-themes**: When auto mode is enabled, users can select:
    - Their preferred **Auto light theme** (from available light themes like Light, Light Contrast, E-Ink).
    - Their preferred **Auto dark theme** (from available dark themes like Dark, Dark Contrast, Solarized).
- **Immediate updates**: Changing the settings applies the theme immediately without requiring an app restart.

## Technical Details
- Updated `Themes.java` to support an `isDark` flag and resolve the theme dynamically based on system `uiMode` and stored preferences.
- Added new string keys and UI preferences in `preferences.xml`.
- Updated `SettingsActivity.java` to populate the filtered dropdowns and handle preference change listeners for immediate UI updates.

## Verification
- Verified that the code compiles successfully with Java 21 and Android SDK 34.
Android

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/data/Settings.java
@@ -448,6 +448,14 @@ public class Settings {
         setString(R.string.pref_key_ui_theme, theme.toString());
     }
 
+    public boolean isAutoThemeEnabled() {
+        return getBoolean(R.string.pref_key_ui_theme_auto, false);
+    }
+
+    public void setAutoThemeEnabled(boolean value) {
+        setBoolean(R.string.pref_key_ui_theme_auto, value);
+    }
+
     public boolean isVolumeButtonsScrollingEnabled() {
         return getBoolean(R.string.pref_key_ui_volumeButtonsScrolling_enabled, false);
     }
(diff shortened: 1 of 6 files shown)


tcitworld:
Thanks, could you rebase your changes?

MarkusLitz:
Done

Strubbl:
I like the automatic switching depending on the system settings. However, I hesitate to approve this because the settings are more complicated now. There are three setting for the theme.

Would it be possible to always enable auto switching of the themes depending on system settings and let the user override the themes in the app? E.g. if the user does not want auto switching the user cannot turn off. But the user can set "dark" theme for both, "auto light theme" and "auto dark theme"?

I am still thinking how we can simplify the options. If you have any idea, please share it.

MarkusLitz:
Yes I agree. However with multiple multiple option for dark and bright I did not see a simple way to do this. 

It would be easier if there would be a designed and marked "Default dark" and "default bright"  -  I guess we could do that, but it would be a separate change we would need to do first.

PR #1493, opened 2026-04-27
Update gradle 9.2.0

Strubbl:
This updates gradle and makes the app work with modern Android (15/16+) edge-to-edge.


tcitworld:
From what I can see on my device this brings a few issues with the header:

image
image
image


Device : Pixel 10 Pro

Strubbl:
Thanks, i did not test with the light theme yesterday. So today i implemented some fixes for that theme. The top status bar and the logo in the nav drawer should be fixed. Also the connection wizard and the settings have been migrated to newer libs and the inset stuff from Android 15/16.
The solarized looks good on my devices.
The only screen which is not updated yet is the About screen. It still uses the old version of the lib. I tried a newer version, but it did not work. There is no documentation for integrating the lib into a non-kotlin, java-based app. Maybe we find an example in another app.

tcitworld:
Thanks for the changes, the same issue is still present on the screen to manage tags for an article. Other than that it seems good.

The issue with the about screen is quite old from what I can remember.

Strubbl:
> the same issue is still present on the screen to manage tags for an article

Good catch, i missed that one. Should also be fixed now.

> The issue with the about screen is quite old from what I can remember.

Yes, it is old.

Strubbl:
i saved my current state of the About screen to my fork repo: https://github.com/Strubbl/wallabag-android-app/pull/4
This needs more work soon™.

tcitworld (approved):

PR #1497, opened 2026-05-13
Disable WebView file access when the image cache is off

### LINKED ISSUE #1496: ReadArticleActivity enables WebView file access on opt-in but never turns it back off
`ReadArticleActivity.initWebView()` does this:

```java
if (settings.isImageCacheEnabled() && !webViewSettings.getAllowFileAccess()) {
    Log.d(TAG, "initWebView() enabling WebView file access");
    webViewSettings.setAllowFileAccess(true);
}
```

It enables `setAllowFileAccess(true)` only when the image cache is on, which on the face of it is sensible. The problem is that on `minSdkVersion 23` the WebView default for `setAllowFileAccess` is already `true` on Android 9 and earlier. So the conditional never explicitly calls `setAllowFileAccess(false)` when the user has the image cache turned off. A WebView created on those Android versions carries the default-on flag regardless of the cache setting.

The article WebView attaches two JS bridges (`hostWebViewTextController` and `hostAnnotationController`) via `addJavascriptInterface(...)`, so any code path that loads a `file://` document into this WebView would let a same-origin `file://` page reach both bridges. CWE-200 maps to the original `setAllowFileAccess(true)` posture.

Suggested fix:

```java
boolean needsFileAccess = settings.isImageCacheEnabled();
if (webViewSettings.getAllowFileAccess() != needsFileAccess) {
    Log.d(TAG, "initWebView() setting WebView file access to " + needsFileAccess);
    webViewSettings.setAllowFileAccess(needsFileAccess);
}
```

Behaviour with the image cache enabled is unchanged. With the image cache off, the WebView is explicitly locked down on every supported Android version, not only API 30+. `loadDataWithBaseURL("file:///android_asset/", ...)` and `file:///android_asset/*` continue to work regardless of the flag.

A PR is open at #1497.

jim-daf:
Closes #1496.

`ReadArticleActivity.initWebView()` already gates `setAllowFileAccess(true)` on the image cache being enabled, but the gate is one-directional:

```java
if (settings.isImageCacheEnabled() && !webViewSettings.getAllowFileAccess()) {
    Log.d(TAG, "initWebView() enabling WebView file access");
    webViewSettings.setAllowFileAccess(true);
}
```

It only enables the flag, it never disables it. On `minSdkVersion 23` the WebView default for `setAllowFileAccess` is `true` on Android 9 and below, so a WebView on those versions kept file URL access regardless of whether the user had the image cache turned on. The article WebView also attaches two JS bridges (`hostWebViewTextController`, `hostAnnotationController`), which is what makes the lingering flag matter.

### Change

Make the flag mirror the cache setting explicitly:

```java
boolean needsFileAccess = settings.isImageCacheEnabled();
if (webViewSettings.getAllowFileAccess() != needsFileAccess) {
    Log.d(TAG, "initWebView() setting WebView file access to " + needsFileAccess);
    webViewSettings.setAllowFileAccess(needsFileAccess);
}
```

### Behaviour

- Image cache on: identical to before (file access on).
- Image cache off: file access is now explicitly off on pre-API-30 devices, not just API 30+.
- `loadDataWithBaseURL("file:///android_asset/", ...)` (the line that loads the article HTML further down) continues to work because the `android_asset` scheme is permitted regardless of this flag.

### DIFF
--- a/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
+++ b/app/src/main/java/fr/gaulupeau/apps/Poche/ui/ReadArticleActivity.java
@@ -646,9 +646,10 @@ public class ReadArticleActivity extends AppCompatActivity {
         WebSettings webViewSettings = webViewContent.getSettings();
         webViewSettings.setJavaScriptEnabled(true);
 
-        if (settings.isImageCacheEnabled() && !webViewSettings.getAllowFileAccess()) {
-            Log.d(TAG, "initWebView() enabling WebView file access");
-            webViewSettings.setAllowFileAccess(true);
+        boolean needsFileAccess = settings.isImageCacheEnabled();
+        if (webViewSettings.getAllowFileAccess() != needsFileAccess) {
+            Log.d(TAG, "initWebView() setting WebView file access to " + needsFileAccess);
+            webViewSettings.setAllowFileAccess(needsFileAccess);
         }
 
         initTtsController();


Strubbl (approved):
Wow, good catch. Thank you for providing this fix.

Issue #1516, opened 2026-06-19
[Android 17] Can't connect over local network


brhahlen:
I've updated my Pixel 9 to Android 17, but have noticed that Wallabag is not syncing, while on the local network.
The error is:

```
Connection test failed
Unknown error: failed to connect to wallabag.domain.com/192.168.x.xx (port 443) from /192.168.y.y (port 409010) after 10000ms
```

When not connected to my local network, everything works fine.
Also, the Wallabagger extension in Vivaldi (/Chrome) works without issue as well.

I _think_ this might be caused by the new permissions that Google added to Android 17, where you need "Local network access" (https://developer.android.com/privacy-and-security/local-network-permission) permission.

hideaki-t:
I have the same issue on an Android 17 device. after some debugging (packet capture/checking logcat, found DNS returned the right A records, but the app even didn't try to connect to the host),
I also end up highly suspecting that Android 17's new "Local network access" permission is the main culprit.

https://github.com/wallabag/android-app/blob/46d125eb0712f5cd02a7b5b2f788fe42b18e75e5/app/build.gradle#L9
local network access will be blocked by default.

I can still connect to the instance via tailscale when needed..

richard7ac:
Same issue here ... if I use VPN when not local it is working ... looks like the Android 17 local access permissions!

Strubbl:
If you want to help testing, there is a debug build apk, which can be installed in parallel to the version you installed from an app store:
https://github.com/wallabag/android-app/pull/1527#issuecomment-5062184750

The test would be to use a wallabag URL which is in the local network and the app should request the local network permission then. This will work on Android 17 only.

Vvalter:
This prevented me from connecting to my local instance on Pixel 8 / Android 17. It took me a couple of hours to figure this out. The final clue was that it connected fine when I enabled tailscale and put the tailscale ip directly, but not when using the local lan ip. A build from this branch works perfectly. Can we get this merged?

PR #1527, opened 2026-07-22
Add permission for local network access and request it at API 37

Matapii:
Related to issue https://github.com/wallabag/android-app/issues/1516

Add permission for local network access and request it at API 37. Tested successfully on Pixel 7. 

Would be happy if the change makes it to master to make the app working again on Android 17.


tcitworld:
Can't we ask for it only if needed if the endpoint is in the local network ?

Strubbl:
I added some code to test if the URL of wallabag is in the local network and only then request that permission. I have no Android 17 device, so it would be nice if someone could test this change.

Matapii:
> I added some code to test if the URL of wallabag is in the local network and only then request that permission. I have no Android 17 device, so it would be nice if someone could test this change.
> 
> If you do not want to build it by yourself, this is my debug build if you want to test this: [app-debug.apk.zip](https://github.com/user-attachments/files/30318999/app-debug.apk.zip)

Thanks for the update. I have tested the changes on an Pixel 7. After installing the app and immediately after the initial launch of it, the app is asking for the permission. I suppose this is not the intended behavior, maybe the default value of the URL in the settings is triggering this behavior. 

However, after granting the permission the app works again on Android 17.

Strubbl:
Yes, thanks for testing. Good point. The URL default is only "https://" which might lead to the permission request trigger (InetAddress.getAllByName("") with empty string resolves to localhost https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#getAllByName-java.lang.String-). I move the check to the end of the configuration, when the config is okay.

Matapii:
I have tested https://github.com/wallabag/android-app/pull/1527#issuecomment-5063438828. Good news is that the app is now starting up without asking for the permission right away. But the app is also not asking for it during the connection setup procedure, so that the original issue occurs. Nevertheless, the permission can be activated manually via android settings.

Strubbl:
Strange, it should ask right after the setup procedure, when the configuration is ok, see [MainActivity.java](https://github.com/wallabag/android-app/pull/1527/changes#diff-7a5b5ea05b600c8a9dd20a9bfcfd232c6eb0edd90a33dfd2c587c5ffc0ce7a87) line 957.

Android 🙈

Strubbl:
I decoupled the check from the connection wizard/config dialog. If it still does not work reliably, maybe you have got an idea?

[app-debug.apk.zip](https://github.com/user-attachments/files/30324961/app-debug.apk.zip)

Matapii:
Added the check into `ConnectionWizardActivity.java‎` and verified that android permission dialog pops up during the connection wizard. After allowing the local connection, the app is able to connect and sync. Pls let me know if this is the correct place or whether there are any other blockers for the PR.

diginfotek:
I have a phone with Android 17. Are there any tests needed?

Currently having this version installed:
https://github.com/wallabag/android-app/pull/1527#issuecomment-5063938245

Matapii:
There is a positive test report here: https://github.com/wallabag/android-app/issues/1516#issuecomment-5309589708

@Strubbl pls let me know if there is anything pending from my side for this PR. Thanks.

csrina:
https://github.com/wallabag/android-app/pull/1527#issuecomment-5063938245 

Tested this on Pixel 8 running Android 17, confirmed it works after manually enabling the permission from the app's android settings page

PR #1531, opened 2026-08-07
Show articles added while offline (#576)

### LINKED ISSUE #576: Display added articles in offline mode
Hi, 
When I'm offline (or online) I often add links from an article I'm reading to read later. But if I'm offline, they won't appear in the list.

Steps to reproduce:
* go offline
* read an article that has a link
* add link to read later
* it will only appear if you go online 

Thanks!

davotoula:
First contribution here — happy to adjust anything to match your conventions.

## Why

Sharing a URL to the app while offline already works: the article is stored locally and an `ADD_LINK` queue item is created. But `ArticleListFragment.getQueryBuilder()` filters on `ArticleId.isNotNull()`, and an unsynced article has no server-side ID — so it appears in no list. You get "Added", then no trace of it until the next sync.

Fixes #576

## What it does

Unsynced articles now appear at the top of Unread, dimmed and italic, showing the URL and "Waiting to sync". They're inert — no tap, no context menu — since there's nothing to read yet and every action needs a server-side ID. On sync they become normal articles.

Screenshot_20260807-095256

Three changes that aren't obvious from the diff:

- `addArticle()` now sets `archive`/`favorite`/`creationDate`. SQLite's `= 0` doesn't match `NULL`, so without this the row stays excluded by the list predicates and sorts last.
- `areItemsTheSame()` used `getArticleId()`, which NPEs on a pending row. It now prefers the server ID and falls back to the local PK only when there isn't one — the PK alone won't do, since a `FULL` update deletes and re-inserts every article.
- `addArticle()` now posts an `ArticlesChangedEvent` for the main feed. Previously only `OfflineQueueChangedEvent` was posted, which no list subscribes to, so a visible list never re-queried.

No schema change; `SCHEMA_VERSION` stays 108. Adds the repo's first JVM unit tests (11) and a `testImplementation junit` dependency.

## Testing

- `./gradlew build` (lint included) green; no new lint findings in changed files.
- 11 unit tests covering the item-identity rules and the pending predicate.
- On device (Pixel 9a, Android 17): added 4 URLs offline — each row appeared immediately at the top of Unread, dimmed and italic; tap and long-press inert; scrolled repeatedly with no state leaking onto normal rows; checked all six themes. Re-enabled data and pulled to refresh: all 4 pushed, each replaced its placeholder, no duplicate rows, queue drained to 0.
- Confirmed via logcat that only the unread list re-queries (`updateList() position: 1`, never `updateAllLists`).

## Known limitations

Deliberately out of scope: articles queued before this change keep `NULL` archive/favorite and stay hidden until their next sync; pending rows don't match search (no indexed content) and show under "untagged"; there's no way to cancel a pending URL.

I also spotted a few pre-existing issues while working on this (a `FULL` sync dropping offline flags on pending rows, and `archiveArticle(String, …)` posting no event) — happy to file those separately.


davotoula:
Hello wallabag team, is there any appetite for this feature?

davotoula:
Bump

